PR #2, opened 2017-09-22
Docs

stephenplusplus:


stephenplusplus:
Just want to call out that I didn't include the samples from [fluent.js](https://github.com/googleapis/nodejs-logging/blob/repo-migration/samples/fluent.js), since I'm not sure where they'd go. Should I stuff that in somewhere?

jmdobry:
> Should I stuff that in somewhere?

Nah.

PR #21, opened 2017-10-23
chore: update-deps: gcp-metadata

ofrobots:
Fixes issue with slow startup when running off of GCP. See: https://github.com/stephenplusplus/gcp-metadata/issues/10.


stephenplusplus:
Thank you! Merged in via #20

PR #34, opened 2017-11-07
Remove Winston/Bunyan samples which have moved.

jmdobry:


jmdobry:
where'd the lint error come from?

stephenplusplus:
See #33. Had to update all the repos today.

PR #51, opened 2018-01-04
1.1.2

stephenplusplus:
## Fixes

- (#50): Use the correct environment variable to detect the region a Google Cloud Function is running in.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@google-cloud/logging",
   "description": "Stackdriver Logging Client Library for Node.js",
-  "version": "1.1.1",
+  "version": "1.1.2",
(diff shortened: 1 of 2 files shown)


ofrobots:
I see the tag landed 4hrs ago, but the release hasn't been published on either github or npm. Is this typically a high latency operation, is something wrong with the automation?

alexander-fenster:
There was no tag, just a pull request called "1.1.2". I need to tag it manually to make a release (doing that right now). After it's tagged, it's almost instant.

alexander-fenster:
If your question was about that, then no, bumping a version in package.json does not cause automatic tagging and releasing (but it's a good idea :) )

PR #83, opened 2018-03-07
Extend system test delay.

stephenplusplus:
Tests are failing randomly. I think this value that sets the delay between a write and a read is too small.


stephenplusplus:
An alternative is to loop until _ time passes, but we are at the mercy of the API server's digestion and response time here.

stephenplusplus:
I think we actually used to have this value set at around 30, maybe even 60 seconds, and somewhere along the way it was set down to 3, which is obviously giving us inconsistent results.

PR #98, opened 2018-03-20
Update google-auto-auth

stephenplusplus:


alexander-fenster:
Same here (`samples/package-lock.json`).

PR #102, opened 2018-03-28
Add trace ID if available

### LINKED ISSUE #23: Write correlation-id to log entries when available
_From @JustinBeckwith on May 12, 2017 14:58_

By writing a correlation-id into the logging metadata, users can get request-correlated logs in pantheon.  See https://gist.github.com/JustinBeckwith/43dd04cfbc4551ddd8996a5b803da0b0 for details.  

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2304_

stephenplusplus:
Fixes #23 

## To Dos
- [ ] Tests

### DIFF
--- a/src/entry.js
+++ b/src/entry.js
@@ -106,6 +106,18 @@ function Entry(metadata, data) {
 
   this.metadata.insertId = this.metadata.insertId || eventId.new();
 
+  if (!this.metadata.trace && global._google_trace_agent) {
+    // Get the current fully qualified trace ID when available from the
+    // @google-cloud/trace-agent library in the LogEntry.trace field format of:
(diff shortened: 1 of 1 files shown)


stephenplusplus:
@JustinBeckwith @ofrobots Is this what we're looking for?

ofrobots:
That would do it. Do we still need a copy of this code in `logging-winston`, `logging-bunyan`.

I can't find the discussion, but it seems like there was some push back when this was initially proposed here. Instead it ended up getting pushed into the winston and bunyan 'integration' libraries.

/cc @kjin

stephenplusplus:
cc @JustinBeckwith -- do we want this?

PR #127, opened 2018-06-19
Synth script

crwilcox:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass

### DIFF
--- a/smoke-test/logging_service_v2_smoke_test.js
+++ b/smoke-test/logging_service_v2_smoke_test.js
@@ -1,0 +1,48 @@
+// Copyright 2018 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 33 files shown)


smoke-test/logging_service_v2_smoke_test.js
```diff
@@ -0,0 +1,48 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+'use strict';
+
+describe('LoggingServiceV2SmokeTest', () => {
+  if (!process.env.SMOKE_TEST_PROJECT) {
```
stephenplusplus: I don't think we set this env var in CI. In the early days of the generator, I think it was intended to switch to the "system test" language, so this type of test would live in `system-test/v2/logging_service.js` for example.

PR #131, opened 2018-06-22
refactor: use mocha for samples

JustinBeckwith:
I'm trying to simplify the universe a little bit.  This PR:
- Removes `repo-tools` from the execution path of npm scripts
- Only submits code coverage during CI, *not* during local runs
- Drops the scaffolding (this will be covered by synthtool)
- Drops ava in favor of mocha, making our tests consistent (only the samples were tested with ava)

If we like all the changes here, we need to figure out how to spread them around to the other repos. Feedback welcomed!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,8 +76,6 @@ jobs:
           command: |-
             mkdir -p /home/node/.npm-global
             npm install
-            npm link
-            chmod +x node_modules/@google-cloud/nodejs-repo-tools/bin/tools
           environment:
             NPM_CONFIG_PREFIX: /home/node/.npm-global
       - run:
@@ -85,7 +83,7 @@ jobs:
           command: npm test
       - run:
           name: Submit coverage data to codecov.
-          command: node_modules/.bin/codecov
+          command: npm run codecov
           when: always
   node8:
     docker:
@@ -109,7 +107,7 @@ jobs:
           name: Link the module being tested to the samples.
           command: |
             cd samples/
-            npm link @google-cloud/logging
+            npm link ../
             npm install
             cd ..
           environment:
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -76,16 +76,14 @@ jobs:
           command: |-
             mkdir -p /home/node/.npm-global
```
alexander-fenster: no need to have `.npm-global` and set `NPM_CONFIG_PREFIX` below if no `npm link`.

samples/package.json
```diff
@@ -6,32 +6,31 @@
...
-    "@google-cloud/logging": "1.2.0",
-    "@google-cloud/logging-bunyan": "0.8.2",
-    "@google-cloud/logging-winston": "0.9.0",
-    "@google-cloud/storage": "1.7.0",
-    "bunyan": "1.8.12",
-    "express": "4.16.3",
-    "fluent-logger": "2.7.0",
-    "winston": "2.4.0",
-    "yargs": "11.0.0"
+    "@google-cloud/logging": "^1.2.0",
```
alexander-fenster: I think carets were not used by design so that samples always used fixed versions, @jmdobry knows more.

JustinBeckwith: That's what lock files are for.

samples/system-test/logs.test.js
```diff
@@ -16,27 +16,26 @@
...
 const cmd = `node logs.js`;
 
 const logName = `nodejs-logging-logs-test-${uuid.v4()}`;
 const message = `Hello world!`;
 
-test.before(tools.checkCredentials);
+before(() => tools.checkCredentials());
 
-test.serial(`should write a log entry`, async t => {
+it(`should write a log entry`, async () => {
```
alexander-fenster: My understanding is that mocha runs tests sequentially. It's OK for this library, but in some others, e.g. spanner or dlp, it will add significant time to the samples tests.

PR #174, opened 2018-07-25
fix: get eslint passing

JustinBeckwith:


JustinBeckwith:
TBR = @googleapis/node-team

PR #193, opened 2018-08-21
fix(gke): correctly detect kubernetes engine

ofrobots:
The value actually returned by google-auth-library is [`KUBERNETES_ENGINE`](https://github.com/google/google-auth-library-nodejs/blob/bad432105f5ad0203266e9c6b31d2235fc96534b/src/auth/envDetect.ts#L21).

This regressed in https://github.com/googleapis/nodejs-logging/pull/128.


ofrobots:
@JustinBeckwith can I get this in a patch release?

PR #196, opened 2018-08-21
Release nodejs-logging v3.0.1

JustinBeckwith:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 [1]: https://www.npmjs.com/package/nodejs-logging?activeTab=versions
 
+## v3.0.1
+
+### Fixes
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,12 @@
 
 [1]: https://www.npmjs.com/package/nodejs-logging?activeTab=versions
 
+## v3.0.1
+
+### Fixes
+- fix(deps): update dependency @google-cloud/logging to v3 (#195)
```
ofrobots: Too late now, but given that this changelog is for the users of this module, this commit is poorly worded. IMO, it should be something like: `fix(samples): upgrade ...`. The way it is worded right now makes it very confusing.

JustinBeckwith: Yeah :(  I need to be more careful about these when they're filed.

PR #203, opened 2018-08-29
chore: upgrade all dependencies and split common

JustinBeckwith:


src/entry.js
```diff
@@ -16,7 +16,7 @@
 
 'use strict';
 
-var commonGrpc = require('@google-cloud/common-grpc');
+var {Service} = require('@google-cloud/common-grpc');
```
fhinkel: const?

src/index.js
```diff
@@ -111,8 +114,6 @@ function Logging(options) {
     return new Logging(options);
   }
 
-  options = common.util.normalizeArguments(this, options);
```
fhinkel: 🎉

src/entry.js
JustinBeckwith: I filed #204 to come back around and update this everywhere 👍

Issue #27, opened 2017-10-24
logging: add zone resource label


stephenplusplus:
_From @ofrobots on September 11, 2017 9:25_

We currently do not detect and attach a `zone` label on the resource. See https://github.com/GoogleCloudPlatform/google-cloud-node/issues/2484#issuecomment-327057968.

_Copied from original issue: GoogleCloudPlatform/google-cloud-node#2601_

PR #215, opened 2018-09-06
fix(GCE): add zone label in GCE descriptor

ofrobots:
Fixes: https://github.com/googleapis/nodejs-logging/issues/27

### DIFF
--- a/src/metadata.js
+++ b/src/metadata.js
@@ -81,6 +81,7 @@ Metadata.getGCEDescriptor = function(callback) {
       // qualified zone name: 'projects/{projectId}/zones/{zone}'. Logging
       // wants just the zone part.
       //
+      // eslint-disable-next-line no-unused-vars
       const [_, zone] =
         zoneResponse.data.match(/projects\/.*\/zones\/(.*)/) || [];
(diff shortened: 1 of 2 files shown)


src/metadata.js
```diff
@@ -75,17 +75,24 @@ Metadata.getGAEDescriptor = function() {
...
-    .instance('id')
-    .then(function(resp) {
+  gcpMetadata.instance('id').then(function(idResponse) {
+    gcpMetadata.instance('zone').then(function(zoneResponse) {
+      // Some parsing is necessary. Metadata service returns a fully
+      // qualified zone name: 'projects/{projectId}/zones/{zone}'. Logging
+      // wants just the zone part.
+      //
+      // eslint-disable-next-line no-unused-vars
+      const [_, zone] =
```
stephenplusplus: This might be more simple, and follows what we usually do in ServiceObject constructors to determine if the user passed us a full resource path, or just the last part:

```js
const zone = zoneResponse.data.split('/').pop();
```

ofrobots: 👍

PR #222, opened 2018-09-14
fix(gce): instance id can be a big number

ofrobots:
Pick up latest `gcp-metadata` that changes the return value on `instance` queries.
In addition this version of `gcp-metadata` fixes silent integer precision loss when
the instance id is too big a number.


ofrobots:
I have manually verified this branch to correctly identify the GCE resource when running on GCE.
image

PR #227, opened 2018-09-16
Use arrow functions

### LINKED ISSUE #206: Use arrow functions

JustinBeckwith:
Fixes #206

### DIFF
--- a/samples/fluent.js
+++ b/samples/fluent.js
@@ -18,7 +18,7 @@
 const express = require('express');
 const app = express();
 
-app.get('*', function(req, res, next) {
+app.get('*', (req, res, next) => {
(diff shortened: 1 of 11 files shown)


samples/fluent.js
```diff
@@ -18,7 +18,7 @@
 const express = require('express');
 const app = express();
 
-app.get('*', function(req, res, next) {
+app.get('*', (req, res, next) => {
   return next('oops');
```
fhinkel: Would it make sense to use short hand notation and leave out {} and return?

JustinBeckwith: Sure!  Will cover in a follow up.

PR #230, opened 2018-09-20
build: write logs to separate file

callmehiphop:
We're seeing some unexpected errors come up in our syslog file. This PR runs tests against a different log file to prevent unexpected results.

@JustinBeckwith I'm not sure if we should continue to dig into the error we're seeing, but it seems to be related to the CI environment and not the client library itself.

```sh
Sep 20 07:44:04 verdaccio puppet-agent[659]: Could not request certificate: Failed to open TCP connection to puppet:8140
```

### DIFF
--- a/system-test/logging.js
+++ b/system-test/logging.js
@@ -281,7 +281,7 @@ describe('Logging', () => {
   });
 
   describe('logs', () => {
-    const log = logging.log('systestlog');
+    const log = logging.log(`system-test-logs-${uuid.v4()}`);
 
     const logEntries = [
       // string data
@@ -314,6 +314,8 @@ describe('Logging', () => {
         },
       },
     };
+
+    after(done => log.delete(done));
 
     it('should list log entries', done => {
       logging.getEntries(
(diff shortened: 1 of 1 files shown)


JustinBeckwith (approved):

JustinBeckwith:
@callmehiphop Yeah... so 2 things there.  

1. Moving to a log specifically for testing logging makes a lot of sense.  That way other things in the project can't interfere with what we're expecting.

2. Yeah, this was me :( I had created a GCE VM with verdaccio on it in our project, and didn't realize that could potentially interfere with our e2e tests.  I have deleted the VM. 

Thank you for digging into this.

system-test/logging.js
```diff
@@ -281,7 +281,7 @@ describe('Logging', () => {
   });
 
   describe('logs', () => {
-    const log = logging.log('syslog');
+    const log = logging.log('systestlog');
```
ofrobots: Can you use a uuid in the string here to make sure two concurrent runs of the tests don't interfere? (yes this has happened!)

ofrobots (approved):
LGTM once comment is addressed.

system-test/logging.js
callmehiphop: Good call! Done

PR #246, opened 2018-10-19
refactor: introduce typescript compiler

ofrobots:
Use TypeScript compiler with the `allowJs` option. With the toolchain
in place we can start migrating the files incrementally.

### DIFF
--- a/src/v2/config_service_v2_client.js
+++ b/src/v2/config_service_v2_client.js
@@ -19,7 +19,7 @@ const gax = require('google-gax');
 const merge = require('lodash.merge');
 const path = require('path');
 
-const VERSION = require('../../package.json').version;
+const VERSION = require('../../../package.json').version;
 
 /**
  * Service for configuring sinks used to export log entries outside of
--- a/src/v2/logging_service_v2_client.js
+++ b/src/v2/logging_service_v2_client.js
@@ -20,7 +20,7 @@ const merge = require('lodash.merge');
 const path = require('path');
 const protobuf = require('protobufjs');
 
-const VERSION = require('../../package.json').version;
+const VERSION = require('../../../package.json').version;
 
 /**
  * Service for ingesting and querying logs.
--- a/src/v2/metrics_service_v2_client.js
+++ b/src/v2/metrics_service_v2_client.js
@@ -19,7 +19,7 @@ const gax = require('google-gax');
 const merge = require('lodash.merge');
 const path = require('path');
 
-const VERSION = require('../../package.json').version;
+const VERSION = require('../../../package.json').version;
 
 /**
  * Service for configuring logs-based metrics.
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,0 +1,24 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "allowJs": true,
+    "declaration": false,
+    "skipLibCheck": true
+  },
+  "include": [
+    "src/*.ts",
+    "src/*/js",
+    "src/**/*.ts",
+    "src/**/*.js",
+    "test/*.ts",
+    "test/*.js",
+    "test/**/*.ts",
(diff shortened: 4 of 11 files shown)


tsconfig.json
```diff
@@ -0,0 +1,20 @@
+{
+  "extends": "./node_modules/gts/tsconfig-google.json",
+  "compilerOptions": {
+    "rootDir": ".",
+    "outDir": "build",
+    "allowJs": true,
+    "declaration": false,
+    "skipLibCheck": true
```
ofrobots: I needed this because I was getting tonnes of type errors in `@google-cloud/common-grpc` type definitions. @JustinBeckwith any ideas?

```
> tsc -p . && cp src/v2/*.json build/src/v2 && cp -r protos build

node_modules/@google-cloud/common-grpc/build/src/service.d.ts:210:5 - error TS2416: Property 'request' in type 'GrpcService' is not assignable to the same property inbase type 'Service'.
  Type '(protoOpts: any, reqOpts: any, callback: any) => any' is not assignable to type '{ (reqOpts: DecorateRequestOptions): Promise<Response>; (reqOpts: DecorateRequestOptions, callback: BodyResponseCallback): void; }'.

210     request(protoOpts: any, reqOpts: any, callback: any): any;
        ~~~~~~~

node_modules/@google-cloud/common-grpc/build/src/service.d.ts:221:5 - error TS2416: Property 'requestStream' in type 'GrpcService' is not assignable to the same property in base type 'Service'.
  Type '(protoOpts: any, reqOpts: any) => any' is not assignable to type '(reqOpts: DecorateRequestOptions) => Request'.

221     requestStream(protoOpts: any, reqOpts: any): any;
```

src/v2/config_service_v2_client.js
```diff
@@ -19,7 +19,7 @@ const gax = require('google-gax');
 const merge = require('lodash.merge');
 const path = require('path');
 
-const VERSION = require('../../package.json').version;
+const VERSION = require('../../../package.json').version;
```
JustinBeckwith: This is auto-generated code :(  It would get replaced the next time we run autosynth.  One option here would be to write some code in synth.py that replaces this line at generation time.  @crwilcox and @kinwa91 would be able to point to a few examples here :)

src/v2/logging_service_v2_client.js
```diff
@@ -20,7 +20,7 @@ const merge = require('lodash.merge');
 const path = require('path');
 const protobuf = require('protobufjs');
 
-const VERSION = require('../../package.json').version;
+const VERSION = require('../../../package.json').version;
```
JustinBeckwith: same as above

src/v2/metrics_service_v2_client.js
```diff
@@ -19,7 +19,7 @@ const gax = require('google-gax');
 const merge = require('lodash.merge');
 const path = require('path');
 
-const VERSION = require('../../package.json').version;
+const VERSION = require('../../../package.json').version;
```
JustinBeckwith: same as above

tsconfig.json
JustinBeckwith: that's weird *scratches head* lemme take a look

src/v2/config_service_v2_client.js
ofrobots: Ideally this code should use [`pkgUp`](https://www.npmjs.com/package/pkg-up) rather than relative paths.

ofrobots: Done.

.eslintrc.yml
```diff
@@ -13,3 +13,4 @@ rules:
   no-warning-comments: warn
   no-var: error
   prefer-const: error
+  node/no-missing-require: warn
```
JustinBeckwith: Why do you need that?  This will also likely be overwritten by synth.

JustinBeckwith:
Approved assuming synth.py runs before, and CI all passes :)

ofrobots:
Can someone confirm if the system tests were passing on kokoro before this PR?

PR #253, opened 2018-10-20
refactor: remove async, methmeth, propprop

JustinBeckwith:
This moves the system tests more towards async/await, while dropping a few dependencies along the way.


system-test/logging.js
```diff
@@ -14,20 +14,15 @@
  * limitations under the License.
  */
 
-'use strict';
```
ofrobots: Is this intentional? What is the rationale?

JustinBeckwith: I am going to run around and do this in all of our TypeScript libs, but we don't need this in our source files.  With the `strict` TypeScript compiler option set, we add `use strict` to all output files by default:
https://github.com/google/ts-style/blob/master/tsconfig-google.json#L14

ofrobots:
Does that apply to js files too?

JustinBeckwith:
Yes! I double checked, and the file in the build folder has `use strict` in it 👍

PR #258, opened 2018-10-23
chore(typescript): convert src/ to typescript

JustinBeckwith:


JustinBeckwith:
Rats.  I went ahead and created tracking issues so we don't step on each other's toes :) . I'll assign stuff to myself before workin on it :)

PR #271, opened 2018-10-24
Fix the synth file and create separate commits

JustinBeckwith:


kinwa91:
I feel that we should not output documents of the underlying gapic generated code - they should only interact with the hand-written layer of the library.

PR #276, opened 2018-10-26
fix(deps): move nock to devDependencies

ofrobots:
Fix regression introduced in a3453de2d1b4715c205af92e8ebf76f0362244dd.

I should have caught it in the code review.


ofrobots:
TBR= @kinwa91 . Making executive decision to land now.

PR #291, opened 2018-10-30
fix(middleware): tweak the middleware api

ofrobots:
* Child suffix is not necessary.
* Make emitRequestLog optional. Make it control finish tracking.

### DIFF
--- a/src/middleware/express/make-middleware.ts
+++ b/src/middleware/express/make-middleware.ts
@@ -32,7 +32,7 @@ export interface AnnotatedRequestType<LoggerType> extends Request {
  * libraries such as winston and bunyan.
  *
  * @param projectId Generated traceIds will be associated with this project.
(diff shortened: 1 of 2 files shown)


src/middleware/express/make-middleware.ts
```diff
@@ -21,17 +21,30 @@ import {Request, Response, NextFunction} from 'express';
...
 }
 
+/**
+ * Generates an express middleware that installs a request-specific logger on
+ * the `request` object. It optionally can do HttpRequest timing that can be
+ * used for generating request logs. This can be used to integrate with logging
+ * libraries such as winston and bunyan.
+ *
+ * @param projectId Generated traceIds will be associated with this project.
+ * @param makeChildLogger A function that generates a logger instances that will
```
kjin:
```suggestion
 * @param makeChildLogger A function that generates logger instances that will
```

PR #297, opened 2018-10-31
fix: fix system tests by choosing semver range for BigQuery

JustinBeckwith:


ofrobots:
lol.. system tests are broken.

ofrobots:
it looks like 'normal flakiness' though. The original problem with BQ require seems fixed.

JustinBeckwith:
aaaaand they're passing again!

PR #298, opened 2018-10-31
fix(metadata): include zone on GAE descriptor

ofrobots:


src/metadata.ts
```diff
@@ -43,12 +51,14 @@ export function getCloudFunctionDescriptor() {
  *
  * @returns {object}
  */
-export function getGAEDescriptor() {
+export async function getGAEDescriptor() {
+  const zone = zoneFromQualifiedZone(await gcpMetadata.instance('zone'));
```
JustinBeckwith: I personally find inline awaits hard to read. Could we split this to two lines?

Issue #133, opened 2018-06-22
Why does this package ship protos


ofrobots:
https://github.com/googleapis/nodejs-logging/blob/01ec04e794966e54450e2b0c7ffd28fab791afea/package.json#L12-L18

Is there a reason why the protos are shipped in the package? @alexander-fenster

crwilcox:
https://github.com/googleapis/nodejs-logging/commit/e01bcd999231aeb3aad6335cda361b39527e49a5#diff-79608869413d6e45d45217ac064e1ad6R91

Because we load them.

JustinBeckwith:
🙋🏼‍♂️ What's the difference between those and the protos we bring in from here?
https://github.com/googleapis/nodejs-proto-files

crwilcox:
That I do not know. @alexander-fenster any info here on that difference? I just know the js code does use those files.

stephenplusplus:
@lukesneeringer was on a mission to destroy `google-proto-files` (if I recall, API teams wanted to more easily test proto changes, and control releases in only one place), then we ended up landing in the middle: google-proto-files still serves all proto files from the [googleapis repo](https://github.com/googleapis/googleapis), and features a loader for "common" proto files (any proto file that is not API-specific, e.g. Date/Time definitions).

ofrobots:
The benefit of statically generated proto bindings is that they can be type-checked by the TypeScript compiler (and you get properly typed interfaces you can use in the code base instead of redefining types). 

How does that play into the roadmap of how we are handling protos?

One example of how we are using static proto codegen is here in the profiler repo: https://github.com/GoogleCloudPlatform/cloud-profiler-nodejs/blob/e038d697f84f1fe302b08b9add76cacdb3d6a049/package.json#L19-L21

ofrobots:
I've changed this from a question to an issue as this requires action IMO. I'm going to remove the 'question' label.

PR #299, opened 2018-10-31
chore: remove a few unused deps

JustinBeckwith:


package.json
```diff
@@ -79,7 +79,6 @@
     "gcp-metadata": "^0.9.0",
     "google-auth-library": "^2.0.0",
     "google-gax": "^0.20.0",
-    "google-proto-files": "^0.17.0",
```
ofrobots: Is this related at all to #133?

JustinBeckwith: It isn't (that I know of). This package is used by gax, but just kinda hanging out here as far as I can tell.

PR #304, opened 2018-11-02
chore(build): use the latest npm on windows for tests

JustinBeckwith:


JustinBeckwith:
I mean I don't know if what I'm doing here makes a difference 😆  but it seemed to work

PR #309, opened 2018-11-07
refactor(ts): improve typescript types

JustinBeckwith:


src/entry.ts
```diff
@@ -19,9 +19,81 @@ import {Service} from '@google-cloud/common-grpc';
 const EventId = require('eventid');
 import * as extend from 'extend';
 import * as is from 'is';
+import {StackdriverHttpRequest} from './http-request';
 
 const eventId = new EventId();
 
+export interface MonitoredResource {
```
ofrobots: What are your thoughts on getting these auto-generated from protobuf? As an example see what we do in the [Stackdriver Profiler agent](https://github.com/googleapis/cloud-profiler-nodejs/tree/master/proto).

To review this, I need to open the protobuf and then verify that you actually did a good job as a protobuf to TS compiler. I would much rather have this auto-generated and correct by construction.

WDYT?

JustinBeckwith: My thought is I like that :) How are y'all generating these?

ofrobots: https://github.com/googleapis/nodejs-logging/pull/314

JustinBeckwith:
ok @ofrobots ready for another look :)

package.json
```diff
@@ -95,7 +95,6 @@
     "eslint-config-prettier": "^3.1.0",
     "eslint-plugin-node": "^8.0.0",
     "eslint-plugin-prettier": "^3.0.0",
-    "google-proto-files": "^0.17.0",
```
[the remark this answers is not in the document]
JustinBeckwith: fixed

JustinBeckwith:
@ofrobots one more time 😆

ofrobots:
The `proto` directory still needs to be copied to `build` and added to the `files` section of `package.json`.

src/log.ts
```diff
@@ -92,8 +125,21 @@ class Log {
...
+  alert(
+      entry: Entry|Entry[],
+      optionsOrCallback?: WriteOptions|ApiResponseCallback,
+      cb?: ApiResponseCallback): void|Promise<ApiResponse> {
+    const options =
+        typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+        typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
+    this.write(
+        Log.assignSeverityToEntries_(entry, 'ALERT'), options, callback!);
```
ofrobots: The non-null assertion on `callback` does not seem correct. When called as the first overload, both `optionsOrCallback` and `cb` can be `undefined`. Why do you need the assertion? I suspect that the function being called has incorrect type signature. Or is this due to the magic promisification-in-the-middle that these libraries use?

src/log.ts
```diff
@@ -124,9 +170,21 @@ class Log {
...
+      entry: Entry|Entry[],
+      optionsOrCallback?: WriteOptions|ApiResponseCallback,
+      cb?: ApiResponseCallback): void|Promise<ApiResponse> {
+    const options =
+        typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+        typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
     const entries = Log.assignSeverityToEntries_(entry, 'CRITICAL');
-    this.write(entries, options, callback);
+    this.write(entries, options, callback!);
```
ofrobots: Same comment as above about the non-null assertion on `callback`.

src/log.ts
```diff
@@ -157,8 +215,21 @@ class Log {
...
+  debug(
+      entry: Entry|Entry[],
+      optionsOrCallback?: WriteOptions|ApiResponseCallback,
+      cb?: ApiResponseCallback): void|Promise<ApiResponse> {
+    const options =
+        typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+        typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
+    this.write(
+        Log.assignSeverityToEntries_(entry, 'DEBUG'), options, callback!);
```
ofrobots: Ditto.

src/log.ts
```diff
@@ -248,9 +323,21 @@ class Log {
...
+      entry: Entry|Entry[],
+      optionsOrCallback?: WriteOptions|ApiResponseCallback,
+      cb?: ApiResponseCallback): void|Promise<ApiResponse> {
+    const options =
+        typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+        typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
     const entries = Log.assignSeverityToEntries_(entry, 'EMERGENCY');
-    this.write(entries, options, callback);
+    this.write(entries, options, callback!);
```
ofrobots: Ditto.

src/log.ts
```diff
@@ -339,8 +426,21 @@ class Log {
...
+  error(
+      entry: Entry|Entry[],
+      optionsOrCallback?: WriteOptions|ApiResponseCallback,
+      cb?: ApiResponseCallback): void|Promise<ApiResponse> {
+    const options =
+        typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+    const callback =
+        typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
+    this.write(
+        Log.assignSeverityToEntries_(entry, 'ERROR'), options, callback!);
```
ofrobots: Ditto.

src/log.ts
```diff
@@ -469,9 +569,16 @@ class Log {
...
+  info(entry: Entry|Entry[], options: WriteOptions, callback: ApiResponseCallback): void;
+  info(entry: Entry|Entry[], callback: ApiResponseCallback): void;
+  info(entry: Entry|Entry[], optionsOrCallback?: WriteOptions|ApiResponseCallback, cb?: ApiResponseCallback):void|Promise<ApiResponse> {
+      const options =
+          typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+      const callback =
+          typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
       this.write(
-          Log.assignSeverityToEntries_(entry, 'INFO'), options, callback);
+          Log.assignSeverityToEntries_(entry, 'INFO'), options, callback!);
```
ofrobots: Ditto.

It seems to me that there is a lot of code being duplicated here. Is there a way to avoid this?

src/log.ts
```diff
@@ -502,9 +609,16 @@ class Log {
...
+  notice(entry: Entry|Entry[], options: WriteOptions, callback: ApiResponseCallback): void;
+  notice(entry: Entry|Entry[], callback: ApiResponseCallback): void;
+  notice(entry: Entry|Entry[], optionsOrCallback?: WriteOptions|ApiResponseCallback, cb?: ApiResponseCallback):void|Promise<ApiResponse> {
+      const options =
+          typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+      const callback =
+          typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
       this.write(
-          Log.assignSeverityToEntries_(entry, 'NOTICE'), options, callback);
+          Log.assignSeverityToEntries_(entry, 'NOTICE'), options, callback!);
```
ofrobots: Ditto.

src/log.ts
```diff
@@ -535,9 +649,16 @@ class Log {
...
+  warning(entry: Entry|Entry[], options: WriteOptions, callback: ApiResponseCallback): void;
+  warning(entry: Entry|Entry[], callback: ApiResponseCallback): void;
+  warning(entry: Entry|Entry[], optionsOrCallback?: WriteOptions|ApiResponseCallback, cb?: ApiResponseCallback):void|Promise<ApiResponse> {
+      const options =
+          typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+      const callback =
+          typeof optionsOrCallback === 'function' ? optionsOrCallback : cb;
       this.write(
-          Log.assignSeverityToEntries_(entry, 'WARNING'), options, callback);
+          Log.assignSeverityToEntries_(entry, 'WARNING'), options, callback!);
```
ofrobots: Ditto.

src/log.ts
```diff
@@ -631,12 +752,16 @@ class Log {
...
    * Another example:
    */
-  write(entry, options?, callback?) {
+  write(entry: Entry|Entry[], options?: WriteOptions): Promise<ApiResponse>;
+  write(entry: Entry|Entry[], options: WriteOptions, callback: ApiResponseCallback): void;
+  write(entry: Entry|Entry[], callback: ApiResponseCallback): void;
+  write(entry: Entry|Entry[], optionsOrCallback?: WriteOptions|ApiResponseCallback, cb?: ApiResponseCallback): void|Promise<ApiResponse> {
+      const options =
+          typeof optionsOrCallback === 'object' ? optionsOrCallback : {};
+      const callback =
```
ofrobots: Suggestion: is this more readable?

```ts
const [options, callback] = 
  typeof optionsOrCallback === 'function' ?
    [{}, optionsOrCallback] : 
    [optionsOrCallback, cb];
```

src/sink.ts
```diff
@@ -17,8 +17,11 @@
 import * as common from '@google-cloud/common-grpc';
 import {promisifyAll} from '@google-cloud/promisify';
 import * as extend from 'extend';
+import {CallOptions} from 'google-gax/build/src/gax';
```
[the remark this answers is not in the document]
JustinBeckwith: 👍 https://github.com/googleapis/gax-nodejs/pull/360

src/log.ts
JustinBeckwith: I tried :( I couldn't figure out a way to define the function signature once, and then effectively "implement" it with different names throughout.  Any ideas?

src/log.ts
JustinBeckwith: That is a lot nicer :) In trying to use it though, I angered the Type gods, and ended up needing to litter more `!` around.  Since this is *such* a common pattern, do you think we should add something like this to common?

```js
  argify<T, C>(optionsOrCallback?: T, cb?: C): [T, C] {
    return typeof optionsOrCallback === 'function' ?
      [{} as T, optionsOrCallback] :
      [optionsOrCallback as T, cb];
  }
```

src/log.ts
ofrobots: Is there an issue tracking the removal of `promisifyAll`?

A casual reader of the code will have the same question as me. This needs to be documented into a TODO in the code until this is cleaned up by the removal of `promisifyAll`.

src/log.ts
ofrobots: I can't think of a way to do this that TypeScript is going to be happy with. Oh well.

src/log.ts
ofrobots: 💖.

Although `argify` is not a great name. How about `maybeOptionsOrCallback`?

src/entry.ts
```diff
@@ -132,7 +149,7 @@ class Entry {
       entry.textPayload = this.data;
     }
     if (is.date(entry.timestamp)) {
-      const seconds = entry.timestamp.getTime() / 1000;
+      const seconds = (this.metadata.timestamp as Date).getTime() / 1000;
```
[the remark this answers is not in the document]
JustinBeckwith: I have no idea why I did this.  I backed it out 🤷‍♂️

src/log.ts
JustinBeckwith: Sure thing.  I created a tracking issue here:
https://github.com/googleapis/google-cloud-node/issues/2887

src/log.ts
JustinBeckwith: SGTM.  PR to implement this over in https://github.com/googleapis/nodejs-common/pull/315 :)

JustinBeckwith:
@ofrobots oooook.  One more time.

ofrobots:
@JustinBeckwith can you address these:

> So, at this point we don't need use proto/logging.js at runtime, and we don't need to ship it (it is rather large).

> Can you drop that file from this PR? You will have to replace the imports of proto/logging to imports of proto/logging.d.

ofrobots:
Awesome. May I suggest this way of doing it instead:

```diff
-"proto:logging": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files -o proto/logging.js google/logging/v2/logging.proto && pbts -o proto/logging.d.ts proto/logging.js"
+"proto:logging": "mkdir -p proto && pbjs -t static-module -w commonjs -p node_modules/google-proto-files google/logging/v2/logging.proto && pbts -o proto/logging.d.ts -"
```

PR #314, opened 2018-11-07
refactor(ts): generate logging types from proto

ofrobots:
Alternative to hand-writing the type-definitions https://github.com/googleapis/nodejs-logging/pull/309.


ofrobots:
The second step would be to check-in the generated `proto/` directory. The proto typedefs will need to be updated very rarely, so it is not necessary to generate them on each CI.

ofrobots:
While we are here, we should also address https://github.com/googleapis/nodejs-logging/issues/133.

PR #338, opened 2018-12-05
fix(docs): const logging = require.. contains binary

kinwa91:
This is preventing docs from generating correctly.


package.json
```diff
@@ -42,7 +42,7 @@
     "check": "gts check",
     "clean": "gts clean",
     "compile": "tsc -p . && cp -r src/v2 build/src/v2 && cp -r protos build && cp test/*.js build/test",
-    "fix": "gts fix && eslint --fix 'samples/*.js' 'samples/**/*.js'",
+    "fix": "gts fix && eslint --fix 'samples/*.js' 'samples/**/*.js' 'src/**/*.js'",
```
JustinBeckwith: Just make this `eslint --fix '**/*.js'`.  If it picks up something it shouldn't we should have that in the `.eslintignore`.

src/v2/.eslintrc.yml
```diff
@@ -0,0 +1,3 @@
+---
```
JustinBeckwith: why?

package.json
JustinBeckwith: I am generally confused.  Why are we not formatting the generated code?

kinwa91: I'm trying to make it format generated code :)

src/v2/.eslintrc.yml
kinwa91: We move the JS generated code under `build/`, which changes where it requires `package.json`.

Without this, `eslint` complains it cannot require `package.json`

package.json
JustinBeckwith: lol sorry, misread.  Regardless, let's try to capture all the js :)

src/v2/.eslintrc.yml
JustinBeckwith: oh yeah!

PR #376, opened 2019-02-01
fix: stop exporting express types publicly

ofrobots:
We don't have a dependency on express. If we expose express types
as part of our public, our dependent modules will need to have a dev
dependency on @types/express. Otherwise they see their builds failing.
Currently there is no way to express this dependency relationship to
npm.

Instead, let's not use express types.

Ref: https://github.com/googleapis/nodejs-logging-bunyan/issues/243


ofrobots:
This doesn't solve the problem generally. It just kicks the can down the road.

PR #400, opened 2019-02-16
feat: ability to detect service context

ofrobots:
Ref: https://github.com/googleapis/nodejs-logging-bunyan/issues/135


ofrobots:
@JustinBeckwith can we get this released?

PR #404, opened 2019-02-25
Logging types async first

### LINKED ISSUE #390: Refactor library to async-first with callbackify
1. Refactor library functions to `async`functions returning promise.
2. Allow optional `callback` that will be picked up by `callbackify`.
3. Refactor functions to use `gax` client directly.

AVaksman:
Fixes #390 and #264 (it's a good idea to open an issue first for discussion)

- [x] Tests and linter pass

1. refactor all methods to async-first.
2. refactor methods to call `gax` client directly.
3. finish types.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,11 +17,12 @@
 import * as common from '@google-cloud/common-grpc';
 import {paginator} from '@google-cloud/paginator';
 import {replaceProjectIdToken} from '@google-cloud/projectify';
-import {promisifyAll} from '@google-cloud/promisify';
+import {callbackifyAll, promisifyAll} from '@google-cloud/promisify';
 import * as arrify from 'arrify';
 import * as extend from 'extend';
 import {GoogleAuth} from 'google-auth-library';
 import * as gax from 'google-gax';
+import {ClientReadableStream} from 'grpc';
 import * as is from 'is';
 
 const pumpify = require('pumpify');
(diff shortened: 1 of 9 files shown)


AVaksman:
For system tests to pass this PR is dependent on paginator PR googleapis/nodejs-paginator/pull/91 PR to land and get a new release.

src/index.ts
```diff
@@ -17,11 +17,12 @@
 import * as common from '@google-cloud/common-grpc';
 import {paginator} from '@google-cloud/paginator';
 import {replaceProjectIdToken} from '@google-cloud/projectify';
-import {promisifyAll} from '@google-cloud/promisify';
+import {callbackifyAll, promisifyAll} from '@google-cloud/promisify';
```
JustinBeckwith: are you still using `promisifyAll` here?

src/index.ts
```diff
@@ -36,19 +37,76 @@ export {HttpRequest};
...
-import {Log, GetEntriesRequest, MonitoredResource, Severity, SeverityNames} from './log';
+import {Entry, LogEntry} from './entry';
+import {Log, GetEntriesRequest, LogOptions, MonitoredResource, Severity, SeverityNames} from './log';
 import {Sink} from './sink';
 import {Duplex} from 'stream';
 import {AbortableDuplex} from '@google-cloud/common';
+import {google} from '../proto/logging';
+import {google as google_config} from '../proto/logging_config';
+import {CallOptions} from 'google-gax/build/src/gax';
+import {Bucket} from '@google-cloud/storage';
```
JustinBeckwith: Are these imported for Types only, or do we actually use the modules?  If types only, please add `// types only` after.  I'm a little worried about the weight of these imports.

src/index.ts
```diff
@@ -36,19 +37,76 @@ export {HttpRequest};
...
-import {Entry} from './entry';
-import {Log, GetEntriesRequest, MonitoredResource, Severity, SeverityNames} from './log';
+import {Entry, LogEntry} from './entry';
+import {Log, GetEntriesRequest, LogOptions, MonitoredResource, Severity, SeverityNames} from './log';
 import {Sink} from './sink';
 import {Duplex} from 'stream';
 import {AbortableDuplex} from '@google-cloud/common';
+import {google} from '../proto/logging';
+import {google as google_config} from '../proto/logging_config';
+import {CallOptions} from 'google-gax/build/src/gax';
```
JustinBeckwith: This *should* be directly exported from `google-gax`.  If it isn't - can you please file a bug over in googleapis/gax-nodejs?

src/index.ts
```diff
@@ -177,6 +237,8 @@ class Logging {
     this.auth = new GoogleAuth(options_);
     this.options = options_;
     this.projectId = this.options.projectId || '{{projectId}}';
+    this.configService = new v2.ConfigServiceV2Client(this.options);
```
JustinBeckwith: This feels like a strange thing to include as part of this review.  Can you explain why we need these to support promise first?

src/index.ts
```diff
@@ -257,7 +319,12 @@ class Logging {
...
    * Another example:
    */
-  createSink(name: string, config, callback) {
+  createSink(name: string, config: CreateSinkRequest): Promise<[Sink, LogSink]>;
+  createSink(
+      name: string, config: CreateSinkRequest,
+      callback: CreateSinkCallback): void;
+  async createSink(name: string, config: CreateSinkRequest):
+      Promise<[Sink, LogSink]> {
     // jscs:enable maximumLineLength
```
JustinBeckwith: Not your code, but can you yank this line?

src/index.ts
```diff
@@ -257,7 +319,12 @@ class Logging {
    * region_tag:logging_create_sink
    * Another example:
    */
-  createSink(name: string, config, callback) {
```
JustinBeckwith: I am really surprised you can add an `async` to the bottom overload like that!

src/index.ts
```diff
@@ -267,38 +334,25 @@ class Logging {
       throw new Error('A sink configuration object must be provided.');
     }
     if (common.util.isCustomType(config.destination, 'bigquery/dataset')) {
-      this.setAclForDataset_(name, config, callback);
-      return;
```
JustinBeckwith: what happened to the return?

src/index.ts
```diff
@@ -267,38 +334,25 @@ class Logging {
...
+      await this.setAclForDataset_(config);
     }
     if (common.util.isCustomType(config.destination, 'pubsub/topic')) {
-      this.setAclForTopic_(name, config, callback);
-      return;
```
JustinBeckwith: return?

src/index.ts
```diff
@@ -267,38 +334,25 @@ class Logging {
...
+      await this.setAclForTopic_(config);
     }
     if (common.util.isCustomType(config.destination, 'storage/bucket')) {
-      this.setAclForBucket_(name, config, callback);
-      return;
```
JustinBeckwith: return?

src/index.ts
```diff
@@ -267,38 +334,25 @@ class Logging {
...
-        (err, resp) => {
-          if (err) {
-            callback(err, null, resp);
-            return;
-          }
-          const sink = self.sink(resp.name);
-          sink.metadata = resp;
-          callback(null, sink, resp);
-        });
+    await this.setProjectId(reqOpts);
```
JustinBeckwith: this looks like a behavior change.  Why aren't we just calling `request` like previously?

AVaksman:
> Ok, so sorry for the annoying feedback here. Can you submit one PR that does just the type changes and d.ts generation, land that, and then come back around with async first? It's hard to tell what's what in this PR cause they're kinda merged together.

I will raise a separate PR for types only first.
Then will follow with async first refactor PRs, perhaps file by file.

src/index.ts
AVaksman: The `request` differentiates whether to make `makeRequestCallback` or `makeRequestStream` depending on whether `callback` is provided or not. 
By converting everything to async-first `callback` serves a flag for action to `callbackify`. As such I am proposing a different approach of methods to call `gax` client directly since `gax` client is available and is not loaded at runtime anymore. 

As I mentioned in the comment below, I will split this PR into multiple smaller PRs.

src/index.ts
AVaksman: I had a concern on how to deal with return signature if a method is flagged `async` => typescript only allows  a `Promie` return type (`Promise<any|void>`) where we need `Promise<any>|void`. I believe it was @callmehiphop who said that I could just use `async` keyword by the implementation but not by overloads.

src/index.ts
AVaksman: These are for types only

src/index.ts
AVaksman: My bad, import from `google-gax` works fine, thanks.

src/index.ts
AVaksman: I will present this in the follow up of smaller PRs

src/index.ts
AVaksman: No, not anymore. Will remove it, thanks.

src/index.ts
AVaksman: Sure thing

JustinBeckwith:
Since we decided to do this differently, going to go ahead and close this out for now.

PR #408, opened 2019-03-01
refactor (typescript): noImplilcitAny

### LINKED ISSUE #264: TypeScript: Enable noImplicitAny
The goal here is to provide perfect TypeScript types for customers using the module.  Generally when doing this work I like to:

1. enable `noImplicitAny` in the tsconfig
2. Choose a single file to get passing, and fix all errors in that file (using VSCode here is really useful)
3. disable`noImplicitAny`
4. Get all the tests passing
5. Submit a PR for just the single file I updated
6. Repeat until finished

It's *a lot* easier if we do a PR per file on this one.

AVaksman:
Fixes #264

- [x] Tests and linter pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,7 +22,9 @@ import * as arrify from 'arrify';
 import * as extend from 'extend';
 import {GoogleAuth} from 'google-auth-library';
 import * as gax from 'google-gax';
+import {ClientReadableStream} from 'grpc';
 import * as is from 'is';
+import * as request from 'request';
 
 const pumpify = require('pumpify');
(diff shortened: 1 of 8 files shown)


src/index.ts
```diff
@@ -22,7 +22,9 @@ import * as arrify from 'arrify';
 import * as extend from 'extend';
 import {GoogleAuth} from 'google-auth-library';
 import * as gax from 'google-gax';
+import {ClientReadableStream} from 'grpc';
 import * as is from 'is';
+import * as request from 'request';
```
JustinBeckwith: This is types only...   right?

PR #409, opened 2019-03-01
refactor: clean up types and imports

JustinBeckwith:
Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)

- [ ] Tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -587,9 +587,7 @@ class Logging {
    *     this.end();
    *   });
    */
-  getEntriesStream(options: GetEntriesRequest) {
-    const self = this;
-    options = options || {};
+  getEntriesStream(options: GetEntriesRequest = {}) {
     let requestStream: Duplex;
     const userStream = streamEvents<Duplex>(pumpify.obj());
     (userStream as AbortableDuplex).abort = () => {
@@ -607,7 +605,7 @@ class Logging {
           },
           options);
       reqOpts.resourceNames = arrify(reqOpts.resourceNames);
-      reqOpts.resourceNames.push('projects/' + self.projectId);
+      reqOpts.resourceNames.push(`projects/${this.projectId}`);
       delete reqOpts.autoPaginate;
       delete reqOpts.gaxOptions;
       const gaxOptions = extend(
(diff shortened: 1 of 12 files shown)


jkwlui:
@stephenplusplus are you able to look into why system tests are failing here?

stephenplusplus:
I can't reproduce locally. Let's try a rebuild!

stephenplusplus:
Rebuild still failed.

@jkwlui-- the error message, "Error: 3 INVALID_ARGUMENT: A monitored resource must be specified for each log entry", could relate to the logic where we find the most appropriate resource to go alongside a log entry dependent on the environment the code is in. That being said, I'm not sure how to simulate the Kokoro environment. Any ideas?

JustinBeckwith:
Can you try something?  If the grpc-js env var is set, just process.exit?  I wonder if this is a race condition with system tests.

stephenplusplus:
I'm not 100% on what that env var is, but I went with this: https://github.com/googleapis/nodejs-logging/pull/409/commits/53691a1676236e48c3dc2ec59ef002cd60a1d8c9

Looks like Kokoro is having issues, though, so the suspense continues!

JustinBeckwith:
@jkwlui is on the other issues

src/index.ts
```diff
@@ -340,10 +339,10 @@ class Logging {
       name: string, config: CreateSinkRequest,
       callback?: CreateSinkCallback): Promise<[Sink, LogSink]>|void {
     const self = this;
-    if (!is.string(name)) {
+    if (typeof name !== 'string') {
```
ofrobots: Would this be considered semver major?

[rant] Removal of `is` as a dependency doesn't fit into the spirit of this PR anyway. This only makes the review process more complex. Refactor PRs should be easy to review.

Open a separate PR for this change.
[/rant]

JustinBeckwith: Happy to :) So question tho - would you prefer a total drop of `is`, a move to the other `is`, or keeping this as `is` 🥁

ofrobots: No objections to dropping `is` as semver-major. If the argument is that it is not breaking, it needs to be made.

PR #412, opened 2019-03-05
refactor: index.ts to async-first

### LINKED ISSUE #390: Refactor library to async-first with callbackify

AVaksman:
Fixes #390 

1. Refactor to async-first with callbackifyAll
    - Currently `logging.request` uses `callback` to differentiate between making `callbackRequest` and 
      `streamRequest'. But now presence of callback will be a flag for `callbackify`

As such
 
2. Refactor methods to call `gax` client methods directly. 
    - To solve above mentioned issue.
    - For readability.


src/index.ts
```diff
@@ -336,49 +340,34 @@ class Logging {
...
-        (err, resp) => {
-          if (err) {
-            callback!(err, null, resp);
-            return;
-          }
-          const sink = self.sink(resp.name);
-          sink.metadata = resp;
-          callback!(null, sink, resp);
-        });
+    await this.setProjectId(reqOpts);
```
stephenplusplus: It's awesome seeing this pattern, where we can get the project ID in an un-obtrusive way with await 👍 I think we have a chance to ditch the `{{projectId}}` placeholder token here altogether. If we just get the project ID before we build the `reqOpts` object (L#360), then we know `this.projectId` is the right value-- no need to use the token for later replacement, since if we can't get it now, we can't get it later, either.

stephenplusplus: Tagging related issue which seeks to destroy the placeholder: https://github.com/googleapis/nodejs-common/issues/10

AVaksman: This method is dependent on objects from other libraries `storage.bucket`, `bigQuery.dataset` or `pubsub.topic` which could still have `{{projectId}}` placeholder. I would have to still run `replaceProjectIdToken` after building the `reqOpts` object. So I placed it after, to have it in one place for both.

JustinBeckwith:
@AVaksman apologies for the merge conflicts on this one...

JustinBeckwith:
Stepping back so @stephenplusplus can do the review on this one :)

src/index.ts
stephenplusplus: Good catch, that is reasonable. I believe this specific method is unique, so I might have picked a bad one to make this suggestion on. For the others, could we skip the `setProjectId()` call, and instead just put the real projectId right where we need it in the reqOpts?

JustinBeckwith:
@AVaksman @stephenplusplus can we pick this back up?  The PR has been sitting around for a while.

src/index.ts
AVaksman: @stephenplusplus 
It looks like I can skip `setProjectId()` `getSinks` and `getSinksStream` since these methods do not depend on the `Sink` constructor.

I am leaving `getEntries` and `getEntriesStream` as is since these guys do depend on the `Log` constructor and I would still need to to `replaceProjectIdToken` for `options.filter` 

https://github.com/googleapis/nodejs-logging/blob/81138d6a5a302b46a8a8bb87a58c320dff1574d0/src/log.ts#L525-L530
from 
https://github.com/googleapis/nodejs-logging/blob/81138d6a5a302b46a8a8bb87a58c320dff1574d0/src/log.ts#L108-L110

AVaksman:
Everything is passing when I am running system tests locally.

AVaksman:
@JustinBeckwith 
To follow up on #429 
Should these imports be removed as well (these libraries are not included in `dependencies`)?
https://github.com/googleapis/nodejs-logging/blob/81138d6a5a302b46a8a8bb87a58c320dff1574d0/src/index.ts#L51-L53

src/index.ts
stephenplusplus: For `getEntries`/`getEntriesStream`, I think we can do the refresh right there:

```ts
this.projectId = await this.auth.getProjectId()
this.formattedName_ = Log.formatName(this.projectId, this.name)
options = extend({ filter: `logName="${this.formattedName_}"` })

// for getEntriesStream
this.auth.getProjectId().then(projectId => {
  this.projectId = projectId
  this.formattedName_ = Log.formatName(this.projectId, this.name)
  options = extend({ filter: `logName="${this.formattedName_}"` })
})
```

The mission for a while has been to eliminate that token altogether, but it was the absence of async/await that has prevented it. But now that we have it, anywhere `this.formattedName_` is used, it should be replaced with code similar to the above, in this file and others where possible. Just a thought, but maybe `formatName()` should be repurposed to do the project ID fetching as well.

src/index.ts
```diff
@@ -683,42 +684,30 @@ class Logging {
...
-      callback?: GetSinksCallback): void|Promise<GetSinksResponse> {
-    const self = this;
-    if (typeof options === 'function') {
-      callback = options as GetSinksCallback;
-      options = {};
-    }
+  async getSinks(opts?: GetSinksRequest|
+                 GetSinksCallback): Promise<GetSinksResponse> {
+    const options = opts ? opts as GetSinksRequest : {};
+    this.projectId = await this.auth.getProjectId();
```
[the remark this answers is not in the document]
AVaksman: In an `async` method (like `getSinks`) I could just add a check 
```ts
if(!this.projectId) {
  this.projectId = awair this.auth.getProjectId();
}
```
In `getSinksStream` method however, that would be
```ts
userStream.once('reading', () => {
  if(!this.projectId) {
    this.auth.getProjectId().then(projectId => {
      this.projectId = projectId;
      const reqOpts = //... rest of the code
  })} else {
        const reqOpts = //... rest of the code
      }
});
```

I was a bit hesitant to duplicate code just like that.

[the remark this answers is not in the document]
stephenplusplus: We also need to check that it's not the token value, so it gets even worse. But! I *think* that the auth client does that. Because, if it's working right, we give the auth client the `projectId` the user gave us. So, the client already knows not to try to detect one, it just returns what the user gave us.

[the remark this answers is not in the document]
AVaksman: wouldn't bet about the auth client, looks like it is just checking for the token

https://github.com/googleapis/nodejs-projectify/blob/0e918a0b3a1f0fc9f99c9d986bb7cf33b3b84c33/src/index.ts#L44-L45

If we add a check for `'undefined'`, then we can make `projectId` in constructor optional and `replaceProjectIdToken` will still pick both (`'{{projectId}}'` or `'undefined'`) up.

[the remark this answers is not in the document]
stephenplusplus: This is the part I was thinking about, that already knows what the provided project ID was: https://github.com/googleapis/google-auth-library-nodejs/blob/8fb65eb913cffa84e0557992f18f75c2a3d2d893/src/auth/googleauth.ts#L161

It will default to user-provided, then start rolling through the different best-guesses. That should mean we don't need to do any prior work, we can just call `this.auth.getProjectId()` and it will return what it should.

AVaksman:
@stephenplusplus PTAL.
In `log.getEntriesStream()` I am setting a "flag" and moving setting the `logName` filter to main `logging.getEntriesStream()` method where I already have the proper `projectId`.

stephenplusplus:
Thanks! I'm okay with that given the tradeoff for simplicity. Instead of making it a pseudo-private option, though, let's just own it and make it a feature of the method and call it `log`. We would accept just the short name, and then we'll format it correctly per the `formatName` method.

AVaksman:
> Thanks! I'm okay with that given the tradeoff for simplicity. Instead of making it a pseudo-private option, though, let's just own it and make it a feature of the method and call it `log`. We would accept just the short name, and then we'll format it correctly per the `formatName` method.

Thanks.
Follow up question. Currently if user passes `filter` to `log.getEntries`/`log.getEntriesStream` or `filter` and `log` to `logging.getEntries`/`logging.getEntriesStream` the `filter` will prevail. Should it be documented or prevented via exception?

AVaksman:
> Follow up question. Currently if user passes `filter` to `log.getEntries`/`log.getEntriesStream` or `filter` and `log` to `logging.getEntries`/`logging.getEntriesStream` the `filter` will prevail. Should it be documented or prevented via exception?

Proposing to construct an advanced filter using `AND` in case both user-provided and logName filters are present.

stephenplusplus:
I agree with that suggestion 👍

AVaksman:
@stephenplusplus 
I pushed my proposed code changes in `72fefc1`
So this PR was intended to introduce `async-first` approach in index.ts only. 
Would you like add changes across the whole repo here?

Is anything else needed for this PR?

P.S.
Do you know why `install.ts` of the `system-test` is failing

bcoe:
@AVaksman just doing some triage, it looks like system tests are failing with the following error:

```
DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

Have you had a chance to look into this at all?

AVaksman:
@bcoe 
somewhere in [`const logging = new Logging();`](https://github.com/googleapis/nodejs-logging/blob/fcd868af12922173970f46a800cc0a73a1e95ef6/system-test/fixtures/sample/src/index.ts#L3) a `util.inspect()` functions is called which throws a deprecation warning [DEP0079](https://nodejs.org/docs/latest-v10.x/api/deprecations.html#deprecations_dep0079_custom_inspection_function_on_objects_via_inspect) which fails the last [test](https://github.com/googleapis/nodejs-logging/blob/master/system-test/install.ts#L44).

Investigating.

stephenplusplus:
> Would you like add changes across the whole repo here?
> Is anything else needed for this PR?

I think if going file by file is easier, that's fine. Thanks for getting all of those changes in!

PR #423, opened 2019-03-14
Release @google-cloud/logging v4.5.0

JustinBeckwith:


JustinBeckwith:
@ofrobots so chances are this is just race conditions with the system tests, since they're getting run twice.  That having been said - we need to get this green before we ship :( Will make this a priority tomorrow.

JustinBeckwith:
Currently blocked on #424

PR #424, opened 2019-03-14
test: reap deleted resources to avoid collisions

JustinBeckwith:
The system tests are trying to delete all buckets, datasets, topics, and sinks that match a given name pattern.  If the system tests run at the same time, this leads to weird errors and conflicts.  This PR tries to fix that :)

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -828,7 +828,8 @@ class Log implements LogSeverityFunctions {
                   this.logging.detectedResource = resource;
                   writeWithResource(resource);
                 },
-                () => {
+                (err) => {
+                  console.log(err)
                   // Ignore errors (the API will speak up if it has an issue).
                   writeWithResource(null);
                 });
--- a/src/log.ts
+++ b/src/log.ts
@@ -829,7 +829,7 @@ class Log implements LogSeverityFunctions {
                   writeWithResource(resource);
                 },
                 (err) => {
-                  console.log(err)
+                  console.log(err);
                   // Ignore errors (the API will speak up if it has an issue).
                   writeWithResource(null);
                 });
--- a/src/log.ts
+++ b/src/log.ts
@@ -822,17 +822,10 @@ class Log implements LogSeverityFunctions {
       } else if (this.logging.detectedResource) {
         writeWithResource(this.logging.detectedResource);
       } else {
-        getDefaultResource(this.logging.auth)
-            .then(
-                (resource) => {
-                  this.logging.detectedResource = resource;
-                  writeWithResource(resource);
-                },
-                (err) => {
-                  console.log(err);
-                  // Ignore errors (the API will speak up if it has an issue).
-                  writeWithResource(null);
-                });
+        getDefaultResource(this.logging.auth).then((resource) => {
+          this.logging.detectedResource = resource;
+          writeWithResource(resource);
+        });
       }
       function writeWithResource(resource: {}|null) {
         let decoratedEntries;
--- a/src/metadata.ts
+++ b/src/metadata.ts
@@ -140,19 +140,23 @@ export function getGlobalDescriptor() {
  */
 export async function getDefaultResource(auth: GoogleAuth) {
   const env = await auth.getEnv();
-  switch (env) {
-    case GCPEnv.KUBERNETES_ENGINE:
-      return getGKEDescriptor();
-    case GCPEnv.APP_ENGINE:
-      return getGAEDescriptor();
-    case GCPEnv.CLOUD_FUNCTIONS:
-      return getCloudFunctionDescriptor();
-    case GCPEnv.COMPUTE_ENGINE:
-      // Test for compute engine should be done after all the rest -
-      // everything runs on top of compute engine.
-      return getGCEDescriptor();
-    default:
-      return getGlobalDescriptor();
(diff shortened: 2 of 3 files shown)


JustinBeckwith:
@stephenplusplus I thought I'd have time to drive this through, but it's getting deep in the weeks :( Can you take this over?  Things that are happening:
- I tried to get clever by reaping resources before the tests start.  So far it seems to work, *except* sinks.  There's some bug where `sink.getMetadata()` isn't returning the creation date like the docs say it should :( 
- Even after reaping resources, I noticed other flakiness caused by eventual consistency with writes.  You just have to wait a really long time to make sure everything is consistent.  
- Because we take the "just wait a really long time" approach to consistency, the tests can take forever, and still be flaky.  I started moving us towards a model where we poll ever second, wait for some sort of sentinal state with the returned entires (I played with properties in the data and with creation timestamps), and using that.
- It looks like log timestamps are in second level precision? so using timestamps to do this work was not really idea.

Basically there's way more than I bargained for here 😆 Can you take this over?  I don't care about the approach taken, I just want the system tests de-flaked.

stephenplusplus:
@JustinBeckwith I got some things under control now, but we're left with the issue about the Logging library not knowing what to use as the default resource for the `ci/kokoro: System test` env. The API refuses to write a log entry without one. So, I'm stalled on that for now :\

JustinBeckwith:
uhhh - @ofrobots or @dazuma any insight on this one?

ofrobots:
Can you just use the Global resource:

```js
{
    type: 'global',
  }
```

stephenplusplus:
The failing test is actually meant to confirm that we default to that in the absence of the user being explicit.

stephenplusplus:
@JustinBeckwith alllllll righty, found a fix, and here's how it went.

We're trying to test that if a user doesn't provide a resource when writing a log entry, we use the global resource. But first, our code tries to figure out the most correct resource based on the environment:

```ts
export async function getDefaultResource(auth: GoogleAuth) {
  const env = await auth.getEnv();
  switch (env) {
    case GCPEnv.KUBERNETES_ENGINE:
      return getGKEDescriptor();
    case GCPEnv.APP_ENGINE:
      return getGAEDescriptor();
    case GCPEnv.CLOUD_FUNCTIONS:
      return getCloudFunctionDescriptor();
    case GCPEnv.COMPUTE_ENGINE:
      // Test for compute engine should be done after all the rest -
      // everything runs on top of compute engine.
      return getGCEDescriptor();
    default:
      return getGlobalDescriptor();
  }
}
```

Kokoro triggers the `GCPEnv.COMPUTE_ENGINE` case, which tries to call out to the metadata server to get the relevant details to build the resource:

```ts
export async function getGCEDescriptor() {
  const idResponse = await gcpMetadata.instance('id');
  // ...
```

That call fails with an `ENOTFOUND`, because, well, we told it to in the system tests:

```ts
// block all attempts to chat with the metadata server (kokoro runs on GCE)
nock('http://metadata.google.internal.')
    .get(() => true)
    .replyWithError({code: 'ENOTFOUND'})
    .persist();
```

The fix: I catch the error and ignore it, just defaulting to the global resource. I also cleaned up the "write to log -> query the log" cycle so we can avoid collisions.

JustinBeckwith:
Oh interesting!  Here's a question though.  

> Kokoro triggers the GCPEnv.COMPUTE_ENGINE case, which tries to call out to the metadata server to get the relevant details to build the resource:

I don't really understand that sentence.  The `nock` call we have in place should prevent `google-auth-library` from ever telling you it's running no GCE.  are you saying this line:

```js
 const env = await auth.getEnv();
```

Is returning COMPUTE_ENGINE?

stephenplusplus:
There's only one way to find out... see you in 10 minutes!

stephenplusplus:
Yep, it does think it is GCE, apparently:

https://source.cloud.google.com/results/invocations/af03624d-164e-4a60-aabb-6833937a5393/targets/cloud-devrel%2Fclient-libraries%2Fnodejs%2Fpresubmit%2Fgoogleapis%2Fnodejs-logging%2Fnode10%2Fsystem-test/log (line 147)

JustinBeckwith:
So I'm not entirely following your fix here, but at a glance:  I'd rather update the test to properly squash the notion we're running in GCE vs making changes to the product code (unless it's the right thing to do).  

I can't imagine how that `nock` call isn't intercepting that `isAvailable` call happening in gcpMetadata. Can you dig deeper?

stephenplusplus:
Welp, this seems to be fixed up and consistently passing after some text fixture refactors.

The source code changes are simply handling the error by defaulting to the global resource, instead of defaulting to no resource. The code comment said something like "We can't find the resource. If that's a problem, the API will speak up". And, it seems like it's always a problem, and the API will always speak up.

I believe  what we want is to default to the global resource not only when the user is not in GAE/GCE/GKE, etc., but when a call to those endpoints fail for any reason as well. If that's not the case, let me know!

JustinBeckwith:
@stephenplusplus I would approve this, but hilariously I can't because it was originally my PR. Do you want me to go ahead and merge?

stephenplusplus:
Haha, woops. Approved and merged!

PR #430, opened 2019-03-19
fix(ts): do not require storage/pubsub types, add install test

### LINKED ISSUE #429: Unsatisfied types dependencies in 4.5.0
I am getting the following error with the latest `4.5.0`:

```
node_modules/@google-cloud/logging/build/src/index.d.ts:32:24 - error TS2307: Cannot find module '@google-cloud/storage'.

32 import { Bucket } from '@google-cloud/storage';
                          ~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@google-cloud/logging/build/src/index.d.ts:33:25 - error TS2307: Cannot find module '@google-cloud/bigquery'.

33 import { Dataset } from '@google-cloud/bigquery';
                           ~~~~~~~~~~~~~~~~~~~~~~~~

node_modules/@google-cloud/logging/build/src/index.d.ts:34:23 - error TS2307: Cannot find module '@google-cloud/pubsub'.

34 import { Topic } from '@google-cloud/pubsub';
                         ~~~~~~~~~~~~~~~~~~~~~~
```

Export of these dependencies seem like a mistake.

JustinBeckwith:
Fixes #429.  This PR includes two changes:
1. An install test that verifies we can install the module and use it's types and
2. A change to use `any` in places where the types were previously used

I don't feel great about this fix.  This another case where I'm starting to wonder if packaging types with the package is actually a good idea.  Thoughts/ideas on the approach welcomed


src/index.ts
```diff
@@ -66,7 +66,9 @@ export type DeleteResponse = google.protobuf.Empty;
 export type LogSink = google_config.logging.v2.ILogSink;
 
 export interface CreateSinkRequest {
-  destination: Bucket|Dataset|Topic|string;
+  // destination: Bucket|Dataset|Topic|string;
+  // tslint:disable-next-line no-any
+  destination: any;
```
bcoe: aha, this is the same thing I did in my local testing; glad that I'm figuring these things out.

system-test/install.ts
```diff
@@ -0,0 +1,56 @@
...
+const stagingPath = stagingDir.name;
+const pkg = require('../../package.json');
+
+describe('📦 pack and install', () => {
+  /**
+   * Create a staging directory with temp fixtures used to test on a fresh
+   * application.
+   */
+  it('should be able to use the d.ts', async () => {
+    await execa('npm', ['pack', '--unsafe-perm']);
```
bcoe: I like this approach to integration testing the TypeScript module, I wonder if this could be be abstracted into a module.

ofrobots: @JustinBeckwith I see that you chose not to use [post-install-check](https://npm.im/post-install-check). I am guessing that there are things in that module that you don't like. Can you work with @DominicKramer and provide feedback to make sure we are serving your needs and making that module better for everyone ❤️.

JustinBeckwith: Great call out.  Feedback left!
https://github.com/google/post-install-check/issues

bcoe: Oh, I wasn't trying to be a :trollface:, I didn't realize this was essentially what `post-install-check` does.

It would be nice to eventually consolidate on a module like `post-install-check` ... I honestly don't love the name `post-install-check` if what it does is check that TypeScript has the appropriate dependencies. I read `post-install-check` and assumed it was something more abstract (a generic way of checking your packages post-install scripts?). I'd advocate a name like `check-installed-types`?

ofrobots: Yep, it is not the best name. We can rename it. How about something like `ts-pack-n-play`?

/cc @googleapis/node-team FYI

src/index.ts
```diff
@@ -66,7 +66,9 @@ export type DeleteResponse = google.protobuf.Empty;
 export type LogSink = google_config.logging.v2.ILogSink;
 
 export interface CreateSinkRequest {
-  destination: Bucket|Dataset|Topic|string;
+  // destination: Bucket|Dataset|Topic|string;
+  // tslint:disable-next-line no-any
```
ofrobots: @JustinBeckwith can you add a comment here indicating why any is being used, and what would be an ideal type definition if it were not for the npm/typescript limitations.

system-test/install.ts
JustinBeckwith: FWIW - I love that name.  @ofrobots wanna open an issue over there to consider the rename?

PR #436, opened 2019-03-25
chore: use new GCF env vars in descriptor

DominicKramer:
The `K_SERVICE` and `GOOGLE_CLOUD_REGION` env vars are the
preferred way of getting function name and region information
going forward.

### DIFF
--- a/test/metadata.ts
+++ b/test/metadata.ts
@@ -98,18 +98,34 @@ describe('metadata', () => {
     const K_SERVICE = 'k-service';
     const GOOGLE_CLOUD_REGION = 'google-cloud-region';
 
-    beforeEach(() => {
-      process.env.FUNCTION_NAME = FUNCTION_NAME;
-      process.env.FUNCTION_REGION = FUNCTION_REGION;
-      delete process.env.K_SERVICE;
-      delete process.env.GOOGLE_CLOUD_REGION;
+    const TARGET_KEYS = [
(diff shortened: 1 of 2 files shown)


callmehiphop (approved):
Is it worth documenting this somewhere in the code? I'm wondering if anyone will get tripped up expecting the old env var to take precedence.

test/metadata.ts
```diff
@@ -95,12 +95,37 @@ describe('metadata', () => {
     const FUNCTION_NAME = 'function-name';
     const FUNCTION_REGION = 'function-region';
 
+    const K_SERVICE = 'k-service';
+    const GOOGLE_CLOUD_REGION = 'google-cloud-region';
+
     beforeEach(() => {
       process.env.FUNCTION_NAME = FUNCTION_NAME;
       process.env.FUNCTION_REGION = FUNCTION_REGION;
+      delete process.env.K_SERVICE;
```
JustinBeckwith: Could we use something like sinon to stub instead?  Worried that running the tests will stomp my current session's environment variables.

JustinBeckwith (changes requested):

DominicKramer:
@callmehiphop I've added code docs.

test/metadata.ts
DominicKramer: Sinon looked more complicated than needed, but I've updated the code to use before/after hooks to restore the parts of `process.env` I am modifying.

JustinBeckwith (approved):

Issue #442, opened 2019-04-02
Wrong types for `Entry`


ofrobots:
Consider the constructor for `Entry`.

```ts
class Entry {
  // ...
  constructor(metadata?: LogEntry, data?: Data) {
    this.metadata = extend(
        {
          timestamp: new Date(),
        },
        metadata);
```

It expects a `LogEntry` typed `metadata` value to be passed in. Historically this value has accepted a `timestamp` property with a JS `Date` value in it. You can see that internally it expects the `timestamp` to be a JavaScript `Date` as well. `logging-bunyan` passes in a JS `Date` as well.

The problem is that `LogEntry` type doesn't accept `Date` as a value for `timestamp`.

ofrobots:
In the same vein, it seems that `severity` field on `LogEntry` has types that are too strict.

The following errors in `nodejs-logging-bunyan`.

```ts
new Entry({ severity: 'INFO'});
```

callmehiphop:
[Oof](https://github.com/googleapis/nodejs-logging/blob/master/proto/logging.d.ts#L1708), looks like maybe we shouldn't be using the generated type there.

PR #448, opened 2019-04-10
fix(types): improve types for LogEntry

ofrobots:
Fixes: https://github.com/googleapis/nodejs-logging/issues/442

@bcoe I know you wanted to look at this bug too.  I had some spare cycles so I went ahead and implemented this. LMK if you have another fix already prepared, or have something more holistic about the types we expose across all of the library. We can close this one in that case.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,8 @@
     "pumpify": "^1.5.1",
     "snakecase-keys": "^2.0.0",
     "stream-events": "^1.0.4",
-    "through2": "^3.0.0"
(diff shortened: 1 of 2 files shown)


ofrobots:
I believe this is semver patch as the types have relaxed. An argument can be made that we should restrict the types to accept only `Date` for `timestamp`. That would be semver major however.

callmehiphop:
Looks like you need an `npm run fix` here

ofrobots:
Ah, yup. Done.

PR #449, opened 2019-04-11
Release @google-cloud/logging v4.5.2

ofrobots:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 [1]: https://www.npmjs.com/package/nodejs-logging?activeTab=versions
 
+## v4.5.2
+
+04-11-2019 14:26 PDT
+
+This release has minor bug fixes:
+
+- fix(types): improve types for LogEntry ([#448](https://github.com/googleapis/nodejs-logging/pull/448))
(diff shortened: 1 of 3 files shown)


ofrobots:
What I am not clear on is why https://github.com/googleapis/nodejs-logging/pull/439 is marked as a `fix`. Isn't it a feature?

JustinBeckwith:
Not sure how to look at that one.  @alexander-fenster any thoughts?

alexander-fenster:
@JustinBeckwith @alexander-fenster re: #439, these added headers are required by the spec for all calls which have variable assignment in `google.api.http` annotation (e.g. `post: "v1/{parent=projects/*}/foo"`) but in most cases, it actually works without those headers. In this sense, we are fixing a bug by making the implementation follow the spec it did not follow before, and it does not change any behavior, so for me callling it a `fix:` (and not a `feat:`) is fine.

ofrobots:
Thanks @alexander-fenster . That context was not apparent from the PR itself. For future PRs, can we get a better description so that it is easier to determine semverity?

JustinBeckwith:
That one is on me.  Right now our tools in this space aren't very great.  There's no way to preserve context from a change in the generator to the changes in the underlying library.  As a result, we often have to guess based on the changed output what happened.  We are trying to find ways to make this better, but it's unlikely to happen until we transition to the micro-generator.

PR #452, opened 2019-04-16
refactor: sink.ts to async-first

AVaksman:
Towards #390

Refactor: `sink.ts` to async-first

- [x] Tests and linter pass

### DIFF
--- a/src/sink.ts
+++ b/src/sink.ts
@@ -153,9 +153,8 @@ class Sink {
   delete(gaxOptions: CallOptions, callback: DeleteCallback): void;
   async delete(gaxOptions?: CallOptions|
                DeleteCallback): Promise<DeleteResponse> {
-    this.logging.projectId = await this.logging.auth.getProjectId();
-    this.formattedName_ =
-        'projects/' + this.logging.projectId + '/sinks/' + this.name;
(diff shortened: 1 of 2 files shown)


AVaksman:
@stephenplusplus Please take a look.

src/sink.ts
```diff
@@ -153,23 +151,16 @@ class Sink {
...
   delete(gaxOptions: CallOptions, callback: DeleteCallback): void;
-  delete(gaxOptions?: CallOptions|DeleteCallback, callback?: DeleteCallback):
-      Promise<DeleteResponse>|void {
-    if (typeof gaxOptions === 'function') {
-      callback = gaxOptions as DeleteCallback;
-      gaxOptions = {};
-    }
+  async delete(gaxOptions?: CallOptions|
+               DeleteCallback): Promise<DeleteResponse> {
+    this.logging.projectId = await this.logging.auth.getProjectId();
```
stephenplusplus: Instead of updating the instance variable here, let's just use a local variable to hold the project ID. I think the pattern should just be to get the project ID from the auth client when needed, instead of relying on `this.logging.projectId`.

AVaksman: Done

PR #461, opened 2019-05-02
build!: upgrade engines field to >=8.10.0

bcoe:
This drops support for all Node.js versions < 8.10.0, please upgrade

BREAKING CHANGE: library now requires Node >= 8.10.0


AndrewMorsillo:
This breaks usage with firebase cloud functions. Firebase cloud functions require engines to be "8" or "10" and won't accept any other values. Due to that, this library fails to compile when attempting to deploy cloud functions `firebase deploy --only functions` with

```
i  functions: preparing functions directory for uploading...

Error: Error occurred while parsing your function triggers.

Error: @grpc/grpc-js only works on Node ^8.13.0 || >=10.10.0
```

Issue #472, opened 2019-05-08
should use req.originalUrl for express requestUrl


hgiasac:
Currently `requestUrl` request data in middleware is using `req.url` variable. In express, `req.url` only returns sub-route. 

```javascript
const router = express.Router();
router.get('/', (req, res) => { ... })

app.use('/users', router);
```

![Console Log](https://i.imgur.com/iAhpsfP.png)

There are many cases that we use nested routes in express for modularization. It is better to use `req.originalUrl` for getting full path.

https://expressjs.com/en/api.html#req.originalUrl

I am ready to make PR. However, I need to verify that this feature can be doable or not.

Thanks

bcoe:
@ofrobots can speak to this a bit better than me, but sounds like a worthwhile patch.

ofrobots:
@hgiasac :+1:. Looking forward to the PR!

PR #476, opened 2019-05-09
use originalUrl for Express middleware's request url

hgiasac:
Use `req.originalUrl` instead of `req.url` to get full path (#472 )

- [x] Tests and linter pass

### DIFF
--- a/src/middleware/express/make-http-request.ts
+++ b/src/middleware/express/make-http-request.ts
@@ -18,14 +18,18 @@ import * as http from 'http';
 
 import {StackdriverHttpRequest} from '../../http-request';
 
+export interface ServerRequest extends http.IncomingMessage {
+  originalUrl?: string;
+}
+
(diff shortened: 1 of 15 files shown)


bcoe:
@ofrobots is this looking reasonable to you? seems like a good fix.

src/middleware/express/make-http-request.ts
```diff
@@ -18,14 +18,18 @@ import * as http from 'http';
 
 import {StackdriverHttpRequest} from '../../http-request';
 
+export interface ServerRequest extends http.IncomingMessage {
+  originalUrl: string;
+}
+
 export function makeHttpRequestData(
-  req: http.IncomingMessage,
+  req: ServerRequest,
```
ofrobots: strictly speaking this change is breaking for existing typescript users. However, given that the middleware is marked experimental, this is okay to ship as a non-major.

PR #501, opened 2019-05-31
feat: support apiEndpoint override

JustinBeckwith:
This is going to pass the property onto the underlying gapic generated layer, so no real work is needed :)


package.json
```diff
@@ -75,7 +75,7 @@
   },
   "devDependencies": {
     "@google-cloud/bigquery": "^4.0.0",
-    "@google-cloud/nodejs-repo-tools": "^3.0.0",
+    "@google-cloud/nodejs-repo-tools": "^3.3.0",
```
bcoe: there's an open PR for the repo-metadata.json, at which point we can just get rid of this dependency all together.

JustinBeckwith:
+1 to that - for now I'm just getting the plumbing hooked up

stephenplusplus:
image

PR #528, opened 2019-07-10
fix(docs): add note about batching writes

### LINKED ISSUE #357: Node Event Loop blocking
Hey, I really like the library, but I noted in my metrics that it increases node event loop lag. Example:

```
{
        "stack": [
          "    at new Promise (<anonymous>)", 
          "    at new WrappedPromise (/app/node_modules/async-listener/es6-wrapped-promise.js:9:7)", 
          "    at Promise.then (<anonymous>)", 
          "    at Promise.wrappedThen (/app/node_modules/async-listener/index.js:524:27)", 
          "    at GoogleAuth.getProjectId (/app/node_modules/google-auth-library/build/src/auth/googleauth.js:73:38)", 
          "    at prepareGaxRequest (/app/node_modules/@google-cloud/logging/src/index.ts:741:17)", 
          "    at makeRequestCallback (/app/node_modules/@google-cloud/logging/src/index.ts:765:7)", 
          "    at Logging.request (/app/node_modules/@google-cloud/logging/src/index.ts:738:7)", 
          "    at writeWithResource (/app/node_modules/@google-cloud/logging/src/log.ts:673:22)", 
          "    at metadata_1.getDefaultResource.then (/app/node_modules/@google-cloud/logging/src/log.ts:646:19)", 
          "    at propagateAslWrapper (/app/node_modules/async-listener/index.js:504:23)", 
          "    at /app/node_modules/async-listener/glue.js:188:31", 
          "    at /app/node_modules/async-listener/index.js:541:70", 
          "    at /app/node_modules/async-listener/glue.js:188:31"
        ], 
        "time": 206.7354169921875 // event loop lag
      }
```

is it possible that this library would cause delay in event loop? Didn't check the source code, but if it uses bunyan, or JSON.stringify for logging then it definitely will.

bcoe:
Added note about batching together writes, to address latency concerns with logging.

Fixes #357

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,7 +85,29 @@ async function quickstart(
 }
 
 ```
+# Batching Writes
 
+High throughput applications should consider batching writes, as it introduces latency
+communicating with the upstream service for each request.
+
+```js
+const {Logging} = require('@google-cloud/logging');
+
+// Creates a client
+const logging = new Logging({projectId});
+
+// Selects the log to write to
+const log = logging.log(logName);
+
+// main tain a global array of log lines that have not yet been written.
+const logEntries = []
+logEntries.push({text: 'hello world', timestamp: new Date()});
+
+// we periodically drain the log entries.
+setInterval(async () => {
+  await log.write(logEntries.splice(0));
+}, 500);
+```
 
 
 ## Samples
--- a/README.md
+++ b/README.md
@@ -87,29 +87,22 @@ async function quickstart(
 ```
 ## Batching Writes
 
-High throughput applications should consider batching writes, as it introduces signficant
-latency calling the upstream logging service for each request.
-
-Here's an example of how someone might achieve this:
+High throughput applications should avoid awaiting calls to the logger:
 
 ```js
-const {Logging} = require('@google-cloud/logging');
-
-// Creates a client
-const logging = new Logging({projectId});
-
-// Selects the log to write to
-const log = logging.log(logName);
-
-// maintain a global array of log lines that have not yet been written.
-const logEntries = [];
-logEntries.push({text: 'hello world', timestamp: new Date()});
-
-// we periodically drain the log entries.
-setInterval(async () => {
-  await log.write(logEntries.splice(0));
-}, 500);
+await log.write(logEntry1);
+await log.write(logEntry2);
 ```
+
+Rather, applications should use a _fire and forget_ approach:
+
+```js
+log.write(logEntry1);
+log.write(logEntry2);
+```
+
+The `@google-cloud/logging` library will handle batching and dispatching
+these log lines to the API.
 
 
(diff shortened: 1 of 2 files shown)


steffnay:
Small comment typo lines 24 and 104

README.md
```diff
@@ -85,7 +85,31 @@ async function quickstart(
...
+const log = logging.log(logName);
+
+// maintain a global array of log lines that have not yet been written.
+const logEntries = [];
+logEntries.push({text: 'hello world', timestamp: new Date()});
+
+// we periodically drain the log entries.
+setInterval(async () => {
+  await log.write(logEntries.splice(0));
+}, 500);
```
ofrobots: Google-gax does batching by default, so this may not necessarily be a good default advice. It would be worth documenting how to customize the batching defaults, if they are unsatisfactory. I don't believe we document that presently. /cc @alexander-fenster

bcoe: @alexander-fenster the behavior I'm seeing is that each `await log.write` takes ~`200ms`, which would suggest it's not writing periodically, but waits for the log entry to be completely drained.

I suppose we could advise not awaiting, as an alternative?

alexander-fenster: Gax does not put requests into bundles _by default_ but it's configurable here: https://github.com/googleapis/nodejs-logging/blob/4a1883f9a69ebd3c4d2d3d43561dc83330ef0e02/src/v2/logging_service_v2_client_config.json#L42-L46

Bundling is enabled for `WriteLogEntries` with the following settings:

```
"bundling": {
  "element_count_threshold": 1000,
  "request_byte_threshold": 1048576,
  "delay_threshold_millis": 50
}
```

which means that similar calls (similarity is defined by the _discriminator fields_ in GAPIC config here: https://github.com/googleapis/googleapis/blob/b73d923869aa1d11ea0ec694c2e2ef7a6d4fcea8/google/logging/v2/logging_gapic.yaml#L446-L451) will be bundled into one call, for no longer than 50ms, or until 1000 requests or 1MB of data is queued.

Having that said, it's unclear for me (yet) why `await` takes 200ms.

alexander-fenster: @bcoe Can you try playing with the following options:

```js
const options = {
  gaxOptions: {
    bundleOptions: {
      element_count_threshold: 1000,
      request_byte_threshold: 1048576,
      delay_threshold_millis: 50,
    },
    isBundling: true
  }
};
await log.write(logEntries, options);
```

`isBundling` can disable gax bundling completely, and playing with `delay_threshold_millis` might show how much of those 200ms is caused by gax bundling.

I just hope I figured out the right way of passing these options down to gax via several layers of libraries involved :)

alexander-fenster: Update: there is *no way* of changing the bundling settings (I edited my previous comment).

PR #532, opened 2019-07-22
fix(deps): drop dependency on through2

JustinBeckwith:


bcoe:
@JustinBeckwith I'm leaning towards potentially holding off on the dropping of `through2` for these use cases that are less than trivial, with a tracking issue?

Just want to make sure we're mindful about removing it from repos that were intentional about using its error handling features; what are your thoughts?

PR #551, opened 2019-08-01
chore: update deps

bcoe:
kokoro tests seemed to be hung for a variety of dependencies, so manually updating.

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,10 +53,10 @@
   },
   "dependencies": {
     "@google-cloud/common-grpc": "^1.0.0",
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
Being mildly pedantic ...  shouldn't this . be a `fix(deps): ...`?

bcoe:
@JustinBeckwith what I've been known to do on my own libraries is label a deps update as a `fix` or `feat` if it pulls in anything _that I know of_ that fits this bag, if it's general library maintenance I'm on the fence (can be a bit noisy to show all updates in the CHANGELOG).

PR #573, opened 2019-09-05
fix: add missing function overload

markvandenbrink:
There is a function overload missing in `log.ts` for the function `critical(...)` resulting in a compiler error if `critical` is used as an await function.

- [x] Tests and linter pass


bcoe:
@markvandenbrink good catch 👍 thank you for the contribution.

PR #579, opened 2019-09-16
fix: hide error when writing a log in an undetermined environment

stephenplusplus:


stephenplusplus:
@bcoe just sending this in early for your review. PTAL!

bcoe:
@stephenplusplus I think this should get at the underlying issue:

https://github.com/googleapis/gcp-metadata/pull/239

src/log.ts
```diff
@@ -809,25 +809,39 @@ class Log implements LogSeverityFunctions {
...
+    let decoratedEntries;
+    try {
+      decoratedEntries = self.decorateEntries_(arrify(entry) as Entry[]);
+    } catch (err) {
+      // Ignore errors (the API will speak up if it has an issue).
     }
-    async function writeWithResource(resource: {} | null) {
-      let decoratedEntries;
+
+    const allEntriesHaveResources = decoratedEntries.every(r => {
```
bcoe: I like this approach though, might still be worth considering skipping the metadata check?

stephenplusplus: If `allEntriesHaveResources` turns out to be true, we skip the metadata check below with:

```js
if (allEntriesHaveResources) {
  return writeWithResource(null);
}
```

Is that what you meant?

bcoe:
I think we can probably close

stephenplusplus:
Goodbye, 579. No ticket to Hollywood for you.

PR #606, opened 2019-10-11
refactor: drop dependency on common-grpc

JustinBeckwith:


src/common.ts
```diff
@@ -0,0 +1,217 @@
+/*!
```
stephenplusplus: We should copy over the test suite for these files from common-grpc.

PR #607, opened 2019-10-15
feat: introduce maxEntrySize, for enabling error message truncation

### LINKED ISSUE #427: Error: Attempt to write logs with no transports
_From @dorshay6 on October 11, 2018 13:31_

When sending a large log (body of the requestor sql query) I get the following error and there is no way to limit log size from Winston.

```
[winston] Attempt to write logs with no transports {“message”:“Thu, 11 Oct 2018 13:04:32 GMT uncaughtException: 3 INVALID_ARGUMENT: Log entry with size 318.8K exceeds maximum size of 110.0K, stack: Error: 3 INVALID_ARGUMENT: Log entry with size 318.8K exceeds maximum size of 110.0K\n    at Object.exports.createStatusError (/usr/src/app/node_modules/grpc/src/common.js:87:15)\n    at Object.onReceiveStatus (/usr/src/app/node_modules/grpc/src/client_interceptors.js:1188:28)\n    at InterceptingListener._callNext (/usr/src/app/node_modules/grpc/src/client_interceptors.js:564:42)\n    at InterceptingListener.onReceiveStatus (/usr/src/app/node_modules/grpc/src/client_interceptors.js:614:8)\n    at callback (/usr/src/app/node_modules/grpc/src/client_interceptors.js:841:24)“,”level”:“error”}
```

#### Environment details

  - OS: Docker (node:8.9.4)
  - Node.js version: 8.9.4
  - npm version:
  - `@google-cloud/logging-winston` version: 0.10.2

#### Steps to reproduce

  1. Send a log with over 110k size

_Copied from original issue: googleapis/nodejs-logging-winston#190_

bcoe:
introduces the configuration setting `maxEntrySize` which, if set, will make a best effort to truncate text log entries, such that they don't result in an exception, see:

https://cloud.google.com/logging/quotas

_Note: as follow up, we should set this option to `256,000` on logging-bunyan, and logging-winston._

fixes #427

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -895,16 +895,15 @@ class Log implements LogSeverityFunctions {
       } else {
         // Stackdriver Log Viewer picks up the summary line from the
         // 'message' field.
-        let text: string | null | undefined;
         if (
           entry.jsonPayload &&
           entry.jsonPayload.fields &&
-          entry.jsonPayload.fields.message
+          entry.jsonPayload.fields.message &&
+          entry.jsonPayload.fields.message.stringValue
         ) {
-          text = entry.jsonPayload.fields.message.stringValue;
-        }
-        if (text) {
-          entry!.jsonPayload!.fields!.message.stringValue = text.slice(
+          const text: string | null | undefined =
+            entry.jsonPayload.fields.message.stringValue;
+          entry.jsonPayload.fields.message.stringValue = text.slice(
             0,
             Math.max(text.length - delta, 0)
           );
--- a/src/log.ts
+++ b/src/log.ts
@@ -909,7 +909,6 @@ class Log implements LogSeverityFunctions {
             Math.max(text.length - delta, 0)
           );
         }
-        console.info(JSON.stringify(entry));
       }
     });
   }
--- a/src/log.ts
+++ b/src/log.ts
@@ -24,7 +24,7 @@ import {Response} from 'teeny-request';
 import {google} from '../proto/logging';
 
 import {GetEntriesCallback, GetEntriesResponse, Logging} from '.';
-import {Entry, LogEntry} from './entry';
+import {Entry, EntryJson, LogEntry} from './entry';
 import {getDefaultResource} from './metadata';
 
 const snakeCaseKeys = require('snakecase-keys');
@@ -44,6 +44,7 @@ export interface GetEntriesRequest {
 
 export interface LogOptions {
   removeCircular?: boolean;
+  maxEntrySize?: number; // see: https://cloud.google.com/logging/quotas
 }
 
 export type ApiResponse = [Response];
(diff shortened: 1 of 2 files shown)


src/log.ts
```diff
@@ -866,6 +870,50 @@ class Log implements LogSeverityFunctions {
...
+        ) {
+          text = entry.jsonPayload.fields.message.stringValue;
+        }
+        if (text) {
+          entry!.jsonPayload!.fields!.message.stringValue = text.slice(
+            0,
+            Math.max(text.length - delta, 0)
+          );
+        }
+        console.info(JSON.stringify(entry));
```
JustinBeckwith: Probably don't want to keep that :)

src/log.ts
```diff
@@ -866,6 +870,50 @@ class Log implements LogSeverityFunctions {
...
+   * Truncate log entries at maxEntrySize, so that error is not thrown, see:
+   * https://cloud.google.com/logging/quotas
+   *
+   * @private
+   *
+   * @param {object|string} the JSON log entry.
+   * @returns {object|string} truncated JSON log entry.
+   */
+  private truncateEntries(entries: EntryJson[]) {
+    return entries.forEach(entry => {
```
JustinBeckwith: Dumb question, but does `Array.forEach` actually return anything? Trying to figure out what this method returns.

bcoe: It doesn't return anything; Entries are all serialized objects, which were created by calling `toJSON` on the original set of entries. The `forEach` array is modifying the values in place, before they're passed to the upstream API (I've tested this behavior in `quickstart.js`).

src/log.ts
```diff
@@ -866,6 +870,49 @@ class Log implements LogSeverityFunctions {
...
+        let text: string | null | undefined;
+        if (
+          entry.jsonPayload &&
+          entry.jsonPayload.fields &&
+          entry.jsonPayload.fields.message
+        ) {
+          text = entry.jsonPayload.fields.message.stringValue;
+        }
+        if (text) {
+          entry!.jsonPayload!.fields!.message.stringValue = text.slice(
```
alexander-fenster: If you move this `if (text)` block and this assignment one line up, inside the previous `if`, you won't need to use `!` and the code might even become more straightforward.

src/log.ts
JustinBeckwith: Got it.  Could we drop the `@returns` in the jsdoc, and drop the `return` statement then?  Maybe explicitly add a `:void`?

xtellurian:
What are the units of maxEntrySize?

techieshark:
@xtellurian I assume you sorted this out but for anyone else, it's bytes

https://github.com/googleapis/nodejs-logging-bunyan/blob/34a278c724f9bfcf1511e22b7deb234007f713bf/src/index.ts#L204

PR #609, opened 2019-10-16
fix: truncate additional fields set by winston/bunyan

bcoe:
our Winston and Bunyan loggers set a variety of additional fields that push us above the `256kb` log limit, I've abstracted the approach we use to allow a set of fields to be truncated ... using :stars: magic :stars:.

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -14,9 +14,10 @@
  * limitations under the License.
  */
 
+import arrify = require('arrify');
 import {DeleteCallback} from '@google-cloud/common';
 import {callbackifyAll} from '@google-cloud/promisify';
-import arrify = require('arrify');
+import * as dotProp from 'dot-prop';
 import * as extend from 'extend';
 import {CallOptions} from 'google-gax';
 import {Response} from 'teeny-request';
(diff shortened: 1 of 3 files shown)


src/log.ts
```diff
@@ -886,27 +887,33 @@ class Log implements LogSeverityFunctions {
...
+          // Bunyan:
+          'jsonPayload.fields.msg.stringValue',
+          'jsonPayload.fields.err.structValue.fields.stack.stringValue',
+          'jsonPayload.fields.err.structValue.fields.message.stringValue',
+          // All:
+          'jsonPayload.fields.message.stringValue',
+        ];
+        for (const field of fieldsToTruncate) {
+          const msg: string = dotProp.get(entry, field, '');
+          if (msg !== '') {
```
callmehiphop: Is it possible that `msg` will be undefined? I'm just wondering if we should add a `typeof` check in addition to this.

bcoe: no, the third argument to `dotProp.get` is the default, so it will default to `''` in the case that there's nothing at that path.

bcoe: boop: https://www.npmjs.com/package/dot-prop#getobject-path-defaultvalue

PR #612, opened 2019-10-18
refactor: reduce common dependency

JustinBeckwith:
This just removes a few easy wins.  The only method used by `@google-cloud/common` now is `util.isCustomType`, which will need more specific work to exfiltrate.

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -15,13 +15,10 @@
  */
 
 import arrify = require('arrify');
-import {DeleteCallback} from '@google-cloud/common';
 import {callbackifyAll} from '@google-cloud/promisify';
 import * as dotProp from 'dot-prop';
 import * as extend from 'extend';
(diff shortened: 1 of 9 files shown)


src/log.ts
```diff
@@ -48,10 +45,13 @@ export interface LogOptions {
   maxEntrySize?: number; // see: https://cloud.google.com/logging/quotas
 }
 
-export type ApiResponse = [Response];
+// tslint:disable-next-line no-any
+export type Metadata = any;
+export type ApiResponse = [Metadata];
 export interface ApiResponseCallback {
-  (err: Error | null, apiResponse?: Response): void;
+  (err: Error | null, apiResponse?: Metadata): void;
```
JustinBeckwith: I am not 100% sure what that object is, but I'm 99% sure it ain't going to be a teeny-response.

PR #615, opened 2019-10-19
refactor: improve types towards noImplicitAny

JustinBeckwith:


src/common.ts
```diff
@@ -21,6 +21,10 @@ export interface ObjectToStructConverterConfig {
   stringify?: boolean;
 }
 
+export interface Struct {
```
bcoe: I don't see this `Struct` being used, am I missing something?

JustinBeckwith: You were right!

PR #620, opened 2019-10-25
fix(package): add missing dependency google-auth-library

### LINKED ISSUE #619: google-auth-library is missing from package.json

Please run down the following list and make sure you've tried the usual "quick fixes":

  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

If you are still having issues, please be sure to include as much information as possible:

#### Environment details

  - OS:MacOS
  - Node.js version: 10 and 12
  - npm version:
  - `@google-cloud/logging` version: 5.5.2

#### Steps to reproduce

  1. npm i
  2. Require `@google-cloud/logging`, to be precise https://github.com/elegantmonkeys/good-gcloud/blob/master/lib/index.js#L176

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

AdriVanHoudt:
This got dropped in https://github.com/googleapis/nodejs-logging/commit/51dd05034abbda44b4781fae1a56c835b8cb8605 but it still being used in `src/metadata.ts`

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests and linter pass
- [ ] Code coverage does not decrease (if any source code was changed)

Fixes #619 🦕

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,6 +62,7 @@
     "eventid": "^0.1.2",
     "extend": "^3.0.2",
     "gcp-metadata": "^3.1.0",
+    "google-auth-library": "^5.2.2",
     "google-gax": "^1.7.5",
     "is": "^3.3.0",
     "on-finished": "^2.3.0",


stephenplusplus:
Thank you!

@JustinBeckwith do we have tooling that should have caught this but failed us?

AdriVanHoudt:
@stephenplusplus Thanks for looking into this and merging!
It is indeed odd that the tests didn't catch this as they do `require` metadata as well.

AdriVanHoudt:
@stephenplusplus do you have an ETA on when this will be released? It is kind of ugly to get around this issue atm

stephenplusplus:
Release is on the way after the robots make their way through the processes: #621

JustinBeckwith:
I'm a little surprised this caused trouble to be honest.  Common takes a dependency on auth, so this should be getting pulled in transitively.  Out of curiosity - could you share what's causing trouble?

JustinBeckwith:
And Stephen - oddly I think the answer here is to switch GTS over to using eslint ... Which is possible now!  They have a rule specifically to catch this during linting.

AdriVanHoudt:
@stephenplusplus make the robots work 👌 
@JustinBeckwith regardless of what went wrong I think it bad practice to rely on this behaviour, if you require a package put in package.json and let npm/yarn/... do the deduping/making sure the dep tree works. What I think happens is that npm deduped common and maybe that's why. 

Happens for both node 8 and 12. npm version 6.12.0
```
$ npm ls @google-cloud/common
[project folder]
├─┬ @google-cloud/storage@4.0.0
│ └── @google-cloud/common@2.2.3 
├─┬ @google-cloud/trace-agent@4.2.2
│ └── @google-cloud/common@2.2.3 
├─┬ firebase-admin@8.6.1
│ └─┬ @google-cloud/storage@3.5.0
│   └── @google-cloud/common@2.2.3 
└─┬ good-gcloud@1.0.1
  └─┬ @google-cloud/logging@5.5.2
    └── @google-cloud/common@2.2.3  deduped
```

PR #640, opened 2019-11-17
fix!: properly depend on Long in protos

### LINKED ISSUE #604: Adding this library to a typescript project causes compilations error: `TS2304: Cannot find name 'Long'`
Steps to reproduce:

1. Create a small typescript project that compiles and works
1. In package.json  add `"@google-cloud/logging": "5.4.1"`
1. Import with `import * as googleLogging from '@google-cloud/logging';`
1. Compile again, you will now see errors about missing Long

### WORKAROUND
I did what I describe [here](https://stackoverflow.com/questions/58159115) to work around this issue, but this should not be required or at least should be documented.

**Environment**
```
$ tsc --version
Version 3.5.3
$ node --version
v10.16.2
```

alexander-fenster:
Fixes #604.

We had two sets of `.d.ts` files for protos: one created manually by running `tsc` in `package.json` script, and another created by the code generator. I'm removing the manual ones, and re-generating the generated ones to include an import of type `Long`. `@types/long` is now pulled by `google-gax` so I'm also updating `google-gax` dependency to `^1.11.0` to make sure those types are pulled.

Now, this change is compatible (the code that just uses our 'official' exports will keep working), but I'm changing the contents of the package (see changes to `files` section in `package.json`). It's up to you @bcoe if you want to consider this change breaking.


alexander-fenster:
@bcoe ping on this! are we ready for a major bump here?

Issue #240, opened 2018-10-04
It is totally unclear how to get httpRequest logging work


koldoon:
I'm trying to figure out, how to use API described in the official documentation around httpRequest. For now i've tried to put this data in different parts of LogEntry factory method with no result:

### 1st variant
```
this.log.write( this.log.entry( {
    httpRequest: {
        requestMethod: $req.method,
        requestUrl: $req.url,
        status: res.statusCode
    }
} ) )
```

### 2nd variant:
```
this.log.write( this.log.entry( {
    resource: { type: 'global' },
    severity: this.levelMap.get( entry.level ) || 'DEFAULT',
}, {
    httpRequest: {
        requestMethod: $req.method,
        requestUrl: $req.url,
        status: res.statusCode
    }
} ) )
```

### 3d variant
```
this.log.write( this.log.entry( {
    resource: { type: 'global' },
    severity: this.levelMap.get( entry.level ) || 'DEFAULT',
    httpRequest: {
        requestMethod: $req.method,
        requestUrl: $req.url,
        status: res.statusCode
    }
}, 'comment here' ) )
```

The last one seems to work HALF WAY: in result I get actually two log lines in stackdriver web interface - the first with raw json meta {"resource":{"type":"global"},"severity":"INFO"} and the second what i need.

Please provide clear convenient and documented way to log http requests using your API.

PR #650, opened 2019-12-02
feat(samples): add example of including httpRequest metadata in log

bcoe:
Demonstrates how to include `httpRequest` metadata in a log-entry:

metadata

_Note: I spruced up our code coverage, because I was eyeballing it while I added a test._

fixes: #240


linkinator.config.json
```diff
@@ -2,6 +2,7 @@
   "recurse": true,
   "skip": [
     "https://codecov.io/gh/googleapis/",
-    "www.googleapis.com"
+    "www.googleapis.com",
+    "https://github.com/googleapis/nodejs-logging/blob/master/samples/http-request.js"
```
bcoe: this is an argument for not generating README.md until the nightly run (although I did find an issue with synthtool generating these READMEs https://github.com/googleapis/synthtool/pull/336).

JustinBeckwith: If you have any ideas on how to work around this with linkinator, I'm all ears.  Haven't been able to figure this one out yet :)

package.json
```diff
@@ -30,13 +30,12 @@
     "stackdriver"
   ],
   "scripts": {
-    "cover": "nyc --reporter=lcov mocha build/test/*.js build/test/**/*.js build/test/**/**/*.js && nyc report",
-    "test-no-cover": "mocha build/test/*.js",
-    "test": "npm run cover",
+    "test": "c8 mocha build/test/*.js build/test/**/*.js build/test/**/**/*.js",
```
JustinBeckwith: would `c8 mocha build/test --recurse` be frowned upon?

package.json
```diff
@@ -30,13 +30,12 @@
...
     "docs": "jsdoc -c .jsdoc.js",
     "presystem-test": "npm run compile",
-    "system-test": "mocha build/system-test --timeout 600000",
-    "samples-test": "cd samples/ && npm test && cd ../",
+    "system-test": "c8 mocha build/system-test --timeout 600000",
+    "samples-test": "c8 npm run run-samples-tests",
+    "run-samples-tests": "cd samples/ && npm test && cd ../",
```
JustinBeckwith: This is a new pattern.  If we want to start doing that, should we consider rolling it out broadly and updating the contributing guide?

bcoe: I instead opted for:

```
cd samples/ && c8 npm test && cd ../
```

The same as the original approach, but with the addition of `c8`.

JustinBeckwith: That's great.  I am wondering though - should we consider one hammer PR at this stage to start introducing `c8`?  Maybe cover adding it to all the sample tests at the same time?

linkinator.config.json
bcoe: I think we should potentially land new samples first, and then let the nightly build add the new links... This only bites us when creating brand new libraries.

PR #688, opened 2020-01-28
test: when cleaning, only delete logs that exist

### LINKED ISSUE #550: refactor tests to have less race conditions
Our system tests call `log.delete` as an `after` hook, it's frequently the case that a log line fails to be deleted, even after `>30s` of waiting:

```bash
delete of system-test-logs-df4ebe77-8e09-432b-b971-2adacc2e18c4 failed retries = 1
delete of system-test-logs-df4ebe77-8e09-432b-b971-2adacc2e18c4 failed retries = 2
delete of system-test-logs-df4ebe77-8e09-432b-b971-2adacc2e18c4 failed retries = 3
delete of system-test-logs-a0d900e6-1005-43e0-bf10-b15750367cae failed retries = 1
delete of system-test-logs-a0d900e6-1005-43e0-bf10-b15750367cae failed retries = 2
delete of system-test-logs-a0d900e6-1005-43e0-bf10-b15750367cae failed retries = 3
delete of system-test-logs-dbb4fdef-2234-42b5-acd5-e413bac75c48 failed retries = 1
delete of system-test-logs-dbb4fdef-2234-42b5-acd5-e413bac75c48 failed retries = 2
delete of system-test-logs-dbb4fdef-2234-42b5-acd5-e413bac75c48 failed retries = 3
```

we should dig into this further and, if need be, log an issue with the upstream API.

stephenplusplus:
Fixes #550 (maybe?)

When logs are attempted to be deleted in the after hook, it isn't thoroughly handling errors:

  - Test block (an `it`) gets a reference to a Log (no remote operation, just a Log object)
  - Log object is stored in local array (to the `describe`)
  - Test block tries to create a log entry, but fails
  - After hook iterates over local array of Log objects, but **doesn't know if they were actually created**

This PR just makes sure we're only trying to delete logs that exist.

### DIFF
--- a/system-test/logging.ts
+++ b/system-test/logging.ts
@@ -294,7 +294,10 @@ describe('Logging', () => {
               break;
             }
             retries++;
-            console.warn(`delete of ${log.name} failed retries = ${retries}`, _err.message);
+            console.warn(
+              `delete of ${log.name} failed retries = ${retries}`,
+              _err.message
+            );
             await new Promise(r => setTimeout(r, WRITE_CONSISTENCY_DELAY_MS));
             continue;
           }
--- a/system-test/logging.ts
+++ b/system-test/logging.ts
@@ -290,8 +290,11 @@ describe('Logging', () => {
           try {
             await log.delete();
           } catch (_err) {
+            if (_err.code === 5) {
+              break;
+            }
             retries++;
-            console.warn(`delete of ${log.name} failed retries = ${retries}`);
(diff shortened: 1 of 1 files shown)


bcoe:
I note that system tests pass, which is promising, how does the output look?

stephenplusplus:
Wake up, koko.

stephenplusplus:
No warning output, so clean-up must be working as intended... for now :)

PR #692, opened 2020-01-30
feat: add getLogs() and getLogsStream()

stephenplusplus:
RE: #550

This adds a new method, `logging.getLogs()` and `logging.getLogsStream()` using the logs.list API: https://cloud.google.com/logging/docs/reference/v2/rest/v2/logs/list

### DIFF
--- a/samples/logs.js
+++ b/samples/logs.js
@@ -96,14 +96,14 @@ async function listLogs() {
   // Creates a client
   const logging = new Logging();
 
-  async function listLogs() {
+  async function printLogNames() {
     const [logs] = await logging.getLogs();
     console.log('Logs:');
     logs.forEach(log => {
       console.log(log.name);
     });
   }
-  listLogs();
+  printLogNames();
   // [END logging_list_logs]
 }
 
--- a/samples/logs.js
+++ b/samples/logs.js
@@ -88,6 +88,23 @@ async function writeLogEntryAdvanced(logName, options) {
   // [END logging_write_log_entry_advanced]
 }
 
+async function listLogs() {
+  // [START logging_list_logs]
+  // Imports the Google Cloud client library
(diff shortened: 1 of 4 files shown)


system-test/logging.ts
```diff
@@ -97,6 +98,20 @@ describe('Logging', () => {
       await getAndDelete(logging.getSinks.bind(logging));
     }
 
+    async function deleteLogs() {
+      const [logs] = await logging.getLogs();
```
stephenplusplus: Using `getLogs()` and a naming convention consistent with the other test resources, we can clean up *all* test logs that are matched after each test run.

samples/logs.js
```diff
@@ -88,6 +88,23 @@ async function writeLogEntryAdvanced(logName, options) {
   // [END logging_write_log_entry_advanced]
 }
 
+async function listLogs() {
+  // [START logging_list_logs]
+  // Imports the Google Cloud client library
+  const {Logging} = require('@google-cloud/logging');
```
JustinBeckwith: If this got embedded in cloud.google.com, the code inside of the region tag would have a top level await.  That means folks couldn't reasonably copy and paste the sample without making changes.  Instead, can we follow a model that introduces an async wrapper method?  For Example:

```js
async function listLogsMain() {
  // [START logging_list_logs]
  // Imports the Google Cloud client library
  const {Logging} = require('@google-cloud/logging');
  
  // Creates a client

  async function listLogs() {
    const [logs] = await logging.getLogs();
    console.log('Logs:');
    logs.forEach(log => {
      console.log(log.name);
    });
  }
  listLogs();
  // [END logging_list_logs]
}
```

It does leave a hanging promise, but at least this way the sample is copy/pastable.

PR #693, opened 2020-01-31
docs(samples): move awaits into async scope

stephenplusplus:
The samples were not copy-and-pastable if outside of an async function. This moves any usage of await into a new async function within the sample, then immediately calls it.


stephenplusplus:
system tests failing with "Error: 8 RESOURCE_EXHAUSTED: Quota exceeded for quota metric 'Control requests' and limit 'Control requests per minute' of service 'logging.googleapis.com' for consumer 'project_number:1046198160504'."

PR #721, opened 2020-02-23
chore(test): Ensure writeLogEntries is actually called on log write.

zamnuts:
The test cases for log.write had their assertions within a manual logging service stub for writeLogEntries. If writeLogEntries was never invoked, these tests would still succeed. The refactor for the write test cases use sinon assertion patterns to avoid this scenario. Unit tests for the write suite should now read better, especially as the intention of each test has been refined too.

- [ ] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

~Fixes #<issue_number_goes_here>~ no open issue for this 🦕

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -867,6 +867,7 @@ class Log implements LogSeverityFunctions {
     }
   }
 
+  // TODO proper signature of `private decorateEntries` (sans underscore suffix)
   /**
    * All entries are passed through here in order to get them serialized.
    *
@@ -887,6 +888,7 @@ class Log implements LogSeverityFunctions {
     });
   }
 
+  // TODO consider refactoring `truncateEntries` so that it does not mutate
   /**
    * Truncate log entries at maxEntrySize, so that error is not thrown, see:
    * https://cloud.google.com/logging/quotas
--- a/test/log.ts
+++ b/test/log.ts
@@ -17,6 +17,7 @@ import * as assert from 'assert';
 import {describe, it} from 'mocha';
 import * as extend from 'extend';
 import * as proxyquire from 'proxyquire';
+import * as sinon from 'sinon';
 import {Log as LOG, WriteOptions, GetEntriesRequest} from '../src/log';
 import {Logging} from '../src/index';
 
@@ -341,13 +342,39 @@ describe('Log', () => {
     const ENTRY = {} as Entry;
     const OPTIONS = {};
     const FAKE_RESOURCE = 'fake-resource';
+    let decorateEntriesStub: sinon.SinonStub;
+    let fakeResourceStub: sinon.SinonStub;
+    let projectIdStub: sinon.SinonStub;
+    let writeLogStub: sinon.SinonStub;
+    let origDetectedResource: string;
 
     beforeEach(() => {
-      log.decorateEntries_ = (entries: Entry[]) => entries;
-      fakeMetadata.getDefaultResource = async () => {
-        return FAKE_RESOURCE;
-      };
-      log.logging.auth.getProjectId = async () => PROJECT_ID;
+      decorateEntriesStub = sinon.stub(log, 'decorateEntries_').returnsArg(0);
+      fakeResourceStub = sinon
+        .stub(fakeMetadata, 'getDefaultResource')
+        .resolves(FAKE_RESOURCE);
+      projectIdStub = sinon.stub().resolves(PROJECT_ID);
+      writeLogStub = sinon.stub();
+
+      // loggingService and auth are fixtures defined in the root suite
+      // they don't have these methods, but we need them for this sub suite
+      log.logging.auth.getProjectId = projectIdStub;
+      log.logging.loggingService.writeLogEntries = writeLogStub;
+
+      // primitives used in the coming tests must be cached and swapped
+      // to avoid carrying state to latter tests
+      origDetectedResource = log.logging.detectedResource;
+    });
+
+    afterEach(() => {
+      decorateEntriesStub.restore();
+      fakeResourceStub.restore();
+      log.logging.detectedResource = origDetectedResource;
+    });
+
+    after(() => {
+      delete log.logging.auth.getProjectId;
+      delete log.logging.loggingService.writeLogEntries;
     });
 
     it('should forward options.resource to request', async () => {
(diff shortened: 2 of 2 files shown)


test/log.ts
```diff
@@ -356,46 +383,41 @@ describe('Log', () => {
...
-        throw new Error('Cached resource was not used.');
-      };
-      // tslint:disable-next-line no-any
-      log.logging.loggingService.writeLogEntries = (
-        // tslint:disable-next-line no-any
-        reqOpts: any
-      ) => {
-        assert.strictEqual(reqOpts.resource, log.logging.detectedResource);
-      };
+      const reusableDetectedResource = 'environment-default-resource';
```
stephenplusplus: Appreciate the split out to its own variable to be used in the test, rather than the instance property.

test/log.ts
```diff
@@ -413,161 +435,157 @@ describe('Log', () => {
...
-      };
+      decorateEntriesStub.returns(decoratedEntries);
+      const arrifiedEntries = [ENTRY] as Entry[];
 
       await log.write(ENTRY, OPTIONS);
+      assert(decorateEntriesStub.calledOnceWithExactly(arrifiedEntries));
+      assert(
+        writeLogStub.calledOnceWith(
+          sinon.match({
+            entries: arrifiedEntries,
```
stephenplusplus: I think this should check that what we returned from `decorateEntries_` was used in the API request.

test/log.ts
```diff
@@ -413,161 +435,157 @@ describe('Log', () => {
...
-      const entry = new Entry({}, 'hello world'.padEnd(300000, '.'));
-      // tslint:disable-next-line no-any
-      logger.logging.loggingService.writeLogEntries = (
-        // tslint:disable-next-line no-any
-        reqOpts: any,
-        _gaxOpts: {}
-      ) => {
-        assert.strictEqual(reqOpts.entries[0].textPayload.length, 300000);
-      };
+    describe('truncateEntries', () => {
```
stephenplusplus: While you're in here, could you move this out to a root-level describe? Since it's its own method, it should be tested separately. I would move it from being `private truncateEntries` to just `truncateEntries_` to match our conventions.

No problem if you prefer that work stay out of the scope of this PR.

stephenplusplus:
It looks good! I noticed one side effect, however, in regards to assertion failure output. Here's old vs. new:

```ts
it('should provide helpful output', () => {
  const obj1 = {a: 'b', c: 'd'};
  const obj2 = {a: 'b', c: 'e'};

  try {
    assert.deepStrictEqual(obj1, obj2);
  } catch (e) {
    console.log('assert.deepStrictEqual', e.message);
  }

  try {
    const spy = sinon.spy();
    spy(obj1);
    assert(spy.calledWithExactly(obj2));
  } catch (e) {
    console.log('sinon.spy', e.message);
  }
});
```
```sh
assert.deepStrictEqual Expected values to be strictly deep-equal:
+ actual - expected

  {
    a: 'b',
+   c: 'd'
-   c: 'e'
  }
sinon.spy The expression evaluated to a falsy value:

  assert(spy.calledWithExactly(obj2))
```

Is there something we could use that would provide more helpful messages in the event of a mismatch?

test/log.ts
zamnuts: The original had `decorateEntries_` stubbed out, but I can still do this if you think it's necessary.

test/log.ts
zamnuts: I had thought about it! Depends on our testing philosophy: do we unit test "private" methods (all the things!), or do we test public contracts and allow later refactoring of the implementation (functional)?
I'll look into what other parts of the code rely on truncation, and if it's just used here I think we should leave it as a nested suite.

zamnuts:
I noticed the unhelpful assertion failure output as well. Rewriting using chai would fix it, but would be medium-sized undertaking. Assertion messages would be an alternative, but IMO no more helpful than looking at the `describe`/`it` labels.

test/log.ts
stephenplusplus: Yeah, we ultimately want to know (whether in this test or a separate one) that the decorated entries were used.

stephenplusplus:
@bcoe @JustinBeckwith for thoughts on the output issue!

test/log.ts
stephenplusplus: Our functional tests are just the system tests, otherwise, our tests essentially go function-by-function (every one!), line-by-line. This nested suite is the first deviance I've seen across our libraries (I don't think I reviewed the PR it was introduced in), so it's possible it's slipped in. But, I would prefer we stick to the ease of a straight mapping between function -> test.

test/log.ts
zamnuts: good point. performed this in the "should truncate the entries after decorating" test, and split into 3 separate tests.

src/log.ts
```diff
@@ -877,6 +878,7 @@ class Log implements LogSeverityFunctions {
     });
   }
 
+  // TODO consider refactoring `truncateEntries` so that it does not mutate
```
stephenplusplus: +1

test/log.ts
```diff
@@ -338,16 +343,48 @@ describe('Log', () => {
...
-    const ENTRY = {} as Entry;
-    const OPTIONS = {};
+    let ENTRY: Entry;
+    let ENTRIES: Entry[];
+    let OPTIONS: WriteOptions;
     const FAKE_RESOURCE = 'fake-resource';
+    let truncateEntriesStub: sinon.SinonStub;
+    let decorateEntriesStub: sinon.SinonStub;
+    let fakeResourceStub: sinon.SinonStub;
+    let projectIdStub: sinon.SinonStub;
```
bcoe: I don't love that this adds a lot of logic related to `log` both in the scope of the `describe`, and in the `beforeEach`, I would be tempted to pull this logic into a helper that's something like:

```
function stubLogger(log: Log) {
  // all the mocking logic here.
}
function restoreLogger(log: Log) {

}
```

and I would keep this logic close to:

```js
  before(() => {
    Log = proxyquire('../src/log', {
      '@google-cloud/promisify': fakeCallbackify,
      './entry': {Entry},
      './metadata': fakeMetadata,
    }).Log;
    const assignSeverityToEntries_ = Log.assignSeverityToEntries_;
    Log.assignSeverityToEntries_ = (...args) =>
      (assignSeverityToEntriesOverride || assignSeverityToEntries_).apply(
        null,
        args
      );
  });
```

Sot that it's clear what the heck `log` is, it's a heavily modified Logger instance.

zamnuts: I was trying to enhance the test cases for `write` without refactoring the entire `Log` suite, oh well!

I went ahead and moved the common logger stubs to the root suite. Some of them need to be maintained within their respective sub-suites, e.g. `write`'s stub for `truncateEntries`. I also changed the rest of the tests based on the revised stubbing.

test/log.ts
```diff
@@ -413,439 +398,275 @@ describe('Log', () => {
...
+      'alert',
+      'critical',
+      'debug',
+      'emergency',
+      'error',
+      'info',
+      'notice',
+      'warning',
+    ].forEach(severityMethodName => {
+      describe(severityMethodName, () => {
```
zamnuts: @bcoe are you OK with these dynamic `describe`s? the original tests for the various severity shortcuts all tested the same thing with the only difference being the severity name.

PR #724, opened 2020-02-25
chore(docs): stackdriver logging limits of 256kb/entry and 10mb/write…

### LINKED ISSUE #531: include 256 kb max log line size in documentation
see: https://github.com/googleapis/nodejs-logging/issues/520

zamnuts:
chore(docs): stackdriver logging limits of 256kb/entry and 10mb/write, force compilation before doc script (#531)

- [X] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [X] Ensure the tests and linter pass
- [X] Code coverage does not decrease (if any source code was changed)
- [X] Appropriate docs were updated (if necessary)

Fixes #531 🦕

### DIFF
--- a/package.json
+++ b/package.json
@@ -31,6 +31,7 @@
   "scripts": {
     "test": "c8 mocha --recursive ./build/test",
     "docs": "jsdoc -c .jsdoc.js",
+    "predocs": "npm run compile",
     "presystem-test": "npm run compile",
     "system-test": "c8 mocha build/system-test --timeout 600000",
     "samples-test": "cd samples/ && c8 npm test && cd ../",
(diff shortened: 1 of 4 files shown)


bcoe:
@zamnuts this looks reasonable to me, but we might want to loop in @alexander-fenster and @xiaozhenliu-gg5, as there are plans soon to move all our libraries to TypeScript generation, and we'll want to make sure that we migrate your comments as well.

@xiaozhenliu-gg5 what are the plans for logging?

alexander-fenster:
@bcoe Logging will be one of the two last libraries to convert (the second one is pubsub).

PR #758, opened 2020-03-20
feat!: move API to Typescript generation

xiaozhenliu-gg5:
BREAKING CHANGE: we will bring bundle request back to the client library.


package.json
```diff
@@ -1,19 +1,7 @@
...
-    "node": ">=8.10.0"
-  },
-  "repository": "googleapis/nodejs-logging",
-  "main": "./build/src/index.js",
-  "types": "./build/src/index.d.ts",
-  "files": [
-    "build/src",
-    "build/protos"
-  ],
+  "description": "Stackdriver Logging Client Library for Node.js",
```
zamnuts: stable refactors from the generator would alleviate a lot of diff noise. consider sorting object keys.

see also `src/v2/config_service_v2_client_config.json`'s `non_idempotent` swap

xiaozhenliu-gg5: great suggestion! Will sort package json from microgenerator and pick up the changes here.

xiaozhenliu-gg5:
Can I bother you to take a look at the system-test @bcoe ? the generated system-test fails for not finding the `nodejs-logging` module. (I suspect)
1. I tried to `npm pack` in this repo and created a new clean repo just installs this package and creates a client as `packnplay` does, it throws error of `Cannot find module '@google-cloud/logging'`. 
2. I checked the `npm pack --dry-run` package, it does have all build/src/ files.

xiaozhenliu-gg5:
fixed `system-test` by change package.json to `main: build/src/index.js`  instead of `index.ts`. Thanks @bcoe

PR #778, opened 2020-04-01
feat!: drop node8 support, support for async iterators

alexander-fenster:
BREAKING CHANGE: The library now supports Node.js v10+. The last version to support Node.js v8 is tagged legacy-8 on NPM.

New feature: methods with pagination now support async iteration.


system-test/logging.ts
```diff
@@ -186,6 +189,7 @@ describe('Logging', async () => {
 
     it('should create a sink with a Dataset destination', async () => {
       const sink = logging.sink(generateName());
+      // eslint-disable-next-line @typescript-eslint/no-unused-vars
```
zamnuts, line 192: instead of the eslint disable, could have refactored the next line as `const [, apiResponse] = ...` (omit the `_`)

test/entry.ts
```diff
@@ -37,7 +39,7 @@ const structToObj = (struct: {}) => {
 };
 
 describe('Entry', () => {
-  // tslint:disable-next-line no-any variable-name
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
zamnuts, line 42: what happened to the `variable-name` disable for the cap `E` in `Entry` within a `let`. do the linting rules no longer have this check?

xiaozhenliu-gg5: `linter` does not complain about the variable name, just checked the `gts` package, it might be related with this rule: `"@typescript-eslint/camelcase": "off",`?

system-test/logging.ts
xiaozhenliu-gg5: thanks for the suggestion! Checking the fix out!

test/entry.ts
zamnuts: `variable-name` is now `@typescript-eslint/naming-convention`, see https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/ROADMAP.md#style

I've opened https://github.com/google/gts/issues/486

PR #878, opened 2020-08-17
chore: Add Observability DevX team to logging lib owners

simonz130:


bcoe (approved):

bcoe:
@simonz130 is there an additional step where we need to populate teams? I've added this to a topic for an automation meeting I'm part of tomorrow,  sorry teams are still such a mess.

simonz130:
> @simonz130 is there an additional step where we need to populate teams? I've added this to a topic for an automation meeting I'm part of tomorrow, sorry teams are still such a mess.

We own repos across googleapis and GoogleCloudPlatform and likely it means that we need to have two different teams created that contain the same members in each org and CODEOWNERS should include the right team for the right org. I'm not aware of other steps.

PR #930, opened 2020-11-04
test: make system tests fail accurately locally

freelerobot:
Currently: 
In development mode, we can only `typescript consumer` tests, not the main `logging` tests, bc the tests were in an async function.

Expected: 
Be able to see logging.ts system tests running during development


bcoe (approved):
I might land this commit as:

`test: make system tests fail accurately locally`

or something like that.

`refactor:` will show up as a release in the CHANGELOG, so best to use it for code changes, not test changes.

PR #933, opened 2020-11-06
docs(sample): improve advanced log sample

### LINKED ISSUE #922: sample: improve the advanced log sample in logs.js
Rather than logging "hello world" show a more complex example logging a httprequest type or similar

laljikanjareeya:
We can use Http Request sample for advanced log sample. moved the region tag in http sample for same.

Fixes #922

### DIFF
--- a/samples/http-request.js
+++ b/samples/http-request.js
@@ -29,6 +29,7 @@ async function logHttpRequest(
   responseSize = 256 // response size in bytes.
 ) {
   // [START logging_http_request]
+  // [START logging_write_log_entry_advanced]
   /*
   const projectId = 'YOUR_PROJECT_ID'; // Your Google Cloud Platform project ID
   const logName = 'my-log'; // The name of the log to write to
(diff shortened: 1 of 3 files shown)


nicoleczhu:
Great idea @laljikanjareeya , I can approve after the `mismatched region tag` error is resolved!

laljikanjareeya:
@nicoleczhu It seems like CI issue I have configured the region tag correctly.

JustinBeckwith:
Right now this check is non-blocking, because we're doing some early testing :) @tmatsuo good issue here to check out!

tmatsuo:
There are actually mismatched region tag in logs.js

The following line should be END

https://github.com/googleapis/nodejs-logging/blob/5d3c013bd8390f8a7f41e6432171df3d608bbf33/samples/logs.js#L146

tmatsuo:
@laljikanjareeya Thanks! Now the bot seems happy.

freelerobot (approved):
LGTM. Thank you!

PR #937, opened 2020-11-06
feat: accept entry.timestamp string input in RFC3339 format

### LINKED ISSUE #934: Entry: should let users input custom timestamp string
Given:
```js
const entry = log.entry({
      timestamp: "2020-11-06T20:37:26.466111181Z" // doesn't work
      resource: { type: 'global' },
      severity: 'INFO',
    }, {
      message: 'testing'
    });
```

Handle this more elegantly in the client library

```js
// TODO: change this in entry.ts
if (is.date(entry.timestamp)) {
      const seconds = (entry.timestamp as Date).getTime() / 1000;
      const secondsRounded = Math.floor(seconds);
      entry.timestamp = {
        seconds: secondsRounded,
        nanos: Math.floor((seconds - secondsRounded) * 1e9),
      };
    }
```

freelerobot:
Fixes #934 

Context: 
- Keeps feature parity with [API](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry) 
- Users may find this useful when they need timestamp to be be higher precision than milliseconds, and want to control timestamp input themselves.

### DIFF
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -177,7 +177,7 @@ class Entry {
       const nanoSecs = zuluTime.match(reNano)?.[1];
       entry.timestamp = {
         seconds: ms ? Math.floor(ms / 1000) : 0,
-        nanos: nanoSecs ? parseInt(nanoSecs.padEnd(9, '0')) : 0,
+        nanos: nanoSecs ? Number(nanoSecs.padEnd(9, '0')) : 0,
       };
     }
     return entry;
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -24,7 +24,7 @@ import {objToStruct, structToObj} from './common';
 
 const eventId = new EventId();
 
-export type Timestamp = google.protobuf.ITimestamp | Date;
+export type Timestamp = google.protobuf.ITimestamp | Date | string;
 export type LogSeverity = google.logging.type.LogSeverity | string;
 export type LogEntry = Merge<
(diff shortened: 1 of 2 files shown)


src/entry.ts
```diff
@@ -160,13 +161,24 @@ class Entry {
...
       };
+    } else if (is.string(entry.timestamp)) {
+      // Convert RFC3339 "Zulu" timestamp into a format that can be parsed to Date
+      const zuluTime = entry.timestamp as string;
+      const ms = Date.parse(zuluTime.split(/[.,Z]/)[0] + 'Z');
+      const reNano = /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}.(\d{0,9})Z$/;
+      const nanoSecs = zuluTime.match(reNano)?.[1];
+      entry.timestamp = {
+        seconds: ms ? Math.floor(ms / 1000) : 0,
+        nanos: nanoSecs ? parseInt(nanoSecs.padEnd(9, '0')) : 0,
```
JustinBeckwith, line 180: It's generally safer to use the `Number` constructor instead of `parseInt`.  `parseInt` is more liberal in what it accepts, and in some cases can make assumptions about the radix (leading 0s and octal numbers, etc).

JustinBeckwith (approved):
Nit on usage of `parseInt`, but otherwise LGTM

PR #942, opened 2020-11-10
fix!: users can log httprequest logs without providing a log message

### LINKED ISSUE #929: entry: httprequest without a log message should format correctly
Reported from user issue b/ 172339240

**To repro:** 
```javascript
e = log.entry(httprequest-data, nil)
log.write(e)
```

**yields** 
```JSON
jsonPayload: {
  httpRequest: {
   requestMethod: "get"    
  }
 }
```

**should be**
```
httpRequest: {
   requestMethod: "get"    
}
// jsonPayload is nil
```

This would be a breaking change for users who are used to this behavior, but it would make this client consistent with other languages & product spec

freelerobot:
Fixes #929 🦕

According to specs and [docs](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry), users should be able to log `httpRequest` type logs independently of a log message. 

In this case, the log message (jsonPayload field) can just show up as `{}`. And of course, `jsonPayload` field should not contain the `httpRequest` data.

Note:
- Users can now log `entry(metadata.httprequest, nil)` and jsonPayload will show up as `{}`
- Users can still log `entry(nil, message)`
- Users can still log `entry(metadata, message)` - regular case
- See test cases

Solves customer issue: b/172339240

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -425,10 +425,16 @@ class Log implements LogSeverityFunctions {
    */
   entry(metadataOrData?: LogEntry | string | {}, data?: string | {}) {
     let metadata: LogEntry;
-    if (!data) {
+    if (!data && metadataOrData?.hasOwnProperty('httpRequest')) {
+      // If user logs entry(httpRequest)
+      metadata = metadataOrData as LogEntry;
+      data = {};
+    } else if (!data) {
+      // If user logs entry(message)
       data = metadataOrData as string | {};
       metadata = {};
     } else {
+      // If user logs entry(metadata, message)
       metadata = metadataOrData as LogEntry;
     }
     return this.logging.entry(metadata, data);
--- a/system-test/logging.ts
+++ b/system-test/logging.ts
@@ -534,6 +534,29 @@ describe('Logging', () => {
       });
     });
 
+    it.only('should write a httpRequest log with no message', done => {
+      const {log} = getTestLog();
+
+      const metadata = {
+        httpRequest: { status: 200 }
+      };
+
+      const logEntry = log.entry(metadata);
+
+      log.write(logEntry, err => {
+        assert.ifError(err);
+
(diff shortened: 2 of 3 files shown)


daniel-sanche (approved):

bcoe:
@nicoleczhu this seems like the right fix to me 👍 

The only thing we should be mindful of, is that we try to keep breaking changes to once every year or so. I think we could choose to do a breaking change of the library now, but it might be good to see if there are any other breaking changes we want to make at the same time.

nicoleczhu:
> @nicoleczhu this seems like the right fix to me 👍
> 
> The only thing we should be mindful of, is that we try to keep breaking changes to once every year or so. I think we could choose to do a breaking change of the library now, but it might be good to see if there are any other breaking changes we want to make at the same time.

Makes sense! I was hesitant to release this for that very reason. I'm gonna work on a few more PRs this month, and see if we can bundle a breaking update containing multiple new features/fixes. Tagging this one as `do not merge` for now.

stephenplusplus (approved):

bcoe (approved):

bcoe (commented):
Could we squint, and not call this a breaking change, since I believe we're making it work as intended?

Someone setting `httpRequest` would expect this meta information to be handled differently right?

nicoleczhu:
> Could we squint, and not call this a breaking change, since I believe we're making it work as intended?
> 
> Someone setting `httpRequest` would expect this meta information to be handled differently right?

I think most new users would expect this fix as the default behavior. 

It will only be breaking for users who were aware of the bug on our end && were custom handling the poorly structured logs themselves.

bcoe:
@nicoleczhu it's okay for us to release a major of logging, if you think this is likely to bite people.

bcoe (approved):

PR #954, opened 2020-11-23
ci: fix bucket_not_found race condition in log sink tests

### LINKED ISSUE #953: CI GCP project getting bucket_not_found error
The following log sink in a project you own had errors while exporting logs. Due to this error, logs are not being exported to the sink destination.

Project ID: redacted
Log Export Sink Name
nodejs-logging-sinks-test-55e512d7-3865-45a4-a30a-0b1688252317

Log Export Destination
storage.googleapis.com/nodejs-logging-sinks-test-dba7db92-0812-4a95-bc3d-a114cf7f8686

Error Code
bucket_not_found
Error Detail
The specified storage bucket could not be found by the Cloud Logging service. Create the bucket and grant object create permission for the service account specified in the sink's writerIdentity field on the bucket. You can also set up Cloud Logging to use a different destination.

freelerobot:
Fixes #953 

Scenario that causes this error in CI:
Day 1: Bucket and sink are created
Day 2: Bucket and sink are deleted (this is because the tests do not clean up resources created within the day-of tests, filters for stale resources that are at least 1 hour old)
Day 3: Email error is sent out (this is because day 3 is timed to the next export interval, every 60 minutes)

Buckets can only be deleted at least 1 hour after sink deletion, otherwise:
1 PM: Last log export
1:01 PM: log sink and bucket are deleted
2PM: Logs from 1PM-1:01PM attempts to be exported to bucket (now deleted), which creates error `bucket_not_found` .

**Solution**: 
- In system-test, we delete buckets that are at least 2 days old. 
- In samples-test, we delete buckets that are least 1 hour old (though I may need to extend this to 1-2 days based on how it goes...)

### DIFF
--- a/system-test/logging.ts
+++ b/system-test/logging.ts
@@ -69,17 +69,22 @@ describe('Logging', () => {
     const oneHourAgo = new Date();
     oneHourAgo.setHours(oneHourAgo.getHours() - 1);
 
+    const twoDaysAgo = new Date();
+    twoDaysAgo.setDate(twoDaysAgo.getDate() - 2);
+
     await Promise.all([
+      deleteSinks(),
       deleteBuckets(),
       deleteDatasets(),
       deleteTopics(),
-      deleteSinks(),
     ]);
 
+    // Only delete log buckets that are at least 2 days old
+    // Fixes: https://github.com/googleapis/nodejs-logging/issues/953
     async function deleteBuckets() {
       const [buckets] = await storage.getBuckets({prefix: TESTS_PREFIX});
       const bucketsToDelete = buckets.filter(bucket => {
-        return new Date(bucket.metadata.timeCreated) < oneHourAgo;
+        return new Date(bucket.metadata.timeCreated) < twoDaysAgo;
       });
 
       for (const bucket of bucketsToDelete) {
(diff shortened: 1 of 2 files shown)


daniel-sanche (approved):
LGTM

system-test/logging.ts
```diff
@@ -69,17 +69,22 @@ describe('Logging', () => {
...
       deleteBuckets(),
       deleteDatasets(),
       deleteTopics(),
-      deleteSinks(),
     ]);
 
+    // Only delete log buckets that are at least 2 days old
+    // Fixes: https://github.com/googleapis/nodejs-logging/issues/953
     async function deleteBuckets() {
       const [buckets] = await storage.getBuckets({prefix: TESTS_PREFIX});
```
bcoe, line 85: curious why we chose 1 hour in `sinks.test.js` and  2 days in `logging.ts`?

bcoe (approved):

system-test/logging.ts
nicoleczhu: The sinks.test.js deletes the sink during the same test run (whereas system tests only delete resources older than 1 hour - given our CI schedule this happens 1 day later).

In practice, we'll likely delete the sinks created by sinks.test.js one day later (given our CI schedule).

nicoleczhu: There might still be a race condition here though. I will monitor closely and revise!

PR #955, opened 2020-11-24
perf!: getEntries filters 24 hour timestamp by default

### LINKED ISSUE #924: listlogs: listLogs and listLogEntries should filter by last 24 hours by default

freelerobot:
Fixes #924 

This fix/feat is applied across client libraries in Go, Python, Java and Node, e.g.
https://github.com/googleapis/google-cloud-go/pull/3120
https://github.com/googleapis/python-logging/pull/73

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -462,7 +462,6 @@ class Logging {
     return new Entry(resource, data);
   }
 
-  // TODO: nicole fix this one
   getEntries(options?: GetEntriesRequest): Promise<GetEntriesResponse>;
   getEntries(callback: GetEntriesCallback): void;
   getEntries(options: GetEntriesRequest, callback: GetEntriesCallback): void;
--- a/src/index.ts
+++ b/src/index.ts
@@ -462,6 +462,7 @@ class Logging {
     return new Entry(resource, data);
   }
 
+  // TODO: nicole fix this one
   getEntries(options?: GetEntriesRequest): Promise<GetEntriesResponse>;
   getEntries(callback: GetEntriesCallback): void;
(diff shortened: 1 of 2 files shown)


bcoe (dismissed):
Unless this is changing the API surface, I'd avoid indicating that it's a breaking change with `!`, as this will result in  a major version bump.

nicoleczhu:
@bcoe possible to resolve the change request so this one can be merged as well?

src/index.ts
```diff
@@ -552,12 +552,20 @@ class Logging {
...
-      {
-        orderBy: 'timestamp desc',
-      },
-      options
-    );
+
+    // By default, sort entries by descending timestamp
+    let reqOpts = extend({orderBy: 'timestamp desc'}, options);
+
+    // By default, filter entries to last 24 hours only
```
stephenplusplus, line 559: What was the previous behavior?

nicoleczhu: Previously, it would get as many entries as possible until a max threshold was reached. It took about 10 minutes. The default behavior across CLI is actually to filter by 24 hours, as well as what's recommended in documentation. 
We were getting a lot of support tickets across different language repos on this.
This fix reduces query time to a few seconds.

stephenplusplus: Sounds good, I just wanted to get that information to determine if it's a breaking change, which it sounds like it is. What do you think?

nicoleczhu: Great question, this actually got a lot of debate. After some back and forth in the chats, we decided to mark it as breaking. and bundle it with another breaking fix that just got merged last night

stephenplusplus: Great, thanks!

bcoe (approved):

PR #986, opened 2021-02-04
docs: improve write structured logEntry example

### LINKED ISSUE #973: Improve the Write logEntry sample
Improve the Write logEntry sample on [this page](https://cloud.google.com/logging/docs/samples/logging-write-log-entry)

This page is referenced by [docs](https://cloud.google.com/logging/docs/api/tasks/creating-logs) that explicitly talk about LogEntries. So we should demo this “advanced” logging technique, even though we do recommend the standard log wrapper approach in the quickstarts.

Criteria: 
- [ ] Demo how to write a text logEntry (textPayload)
- [ ] Demo how to write a struct logEntry (jsonPayload)
- [ ] Include: severity level, labels, resource type (“global”)
- [ ] Demo: batch logging vs writing one log at a time
- [ ] Update accompanying tests

Region tag: `logging_write_log_entry`

freelerobot:
Fixes #973  🦕

For internal reference, here's the [canonical sample](https://docs.google.com/document/d/19JMfEL-PiKHe2IDoccz6ZEDPY7BXPDEEqjXgB2-XiSk/edit#)


samples/logs.js
```diff
@@ -16,48 +16,51 @@
...
-  // Creates a client
   const logging = new Logging();
 
   /**
-   * TODO(developer): Uncomment the following line to run the code.
+   * TODO(developer): Uncomment the following line and replace with your values.
    */
-  // const logName = 'Name of the log to write to, e.g. my-log';
-
+      // const logName = 'my-log';
```
JustinBeckwith: If the user needs to define `logName` to run the sample, we like to keep these placeholders around.  They are only going to see what's inside the region tags!

samples/logs.js
```diff
@@ -174,72 +177,72 @@ async function deleteLog(logName) {
...
-      'node $0 write my-log \'{"type":"global"}\' \'{"message":"Hello World!"}\'',
-      'Write a JSON log entry.'
-    )
-    .example('node $0 delete my-log', 'Delete "my-log".')
-    .wrap(120)
-    .recommendCommands()
-    .epilogue('For more information, see https://cloud.google.com/logging/docs')
-    .help()
-    .strict().argv;
+      .demand(1)
```
JustinBeckwith: No need to do it right now, but we want to move away from `yargs` and towards having each snippet in its own file.    BigQuery has some great up to date examples:
https://github.com/googleapis/nodejs-bigquery/tree/master/samples

samples/logs.js
```diff
@@ -16,48 +16,51 @@
...
+    name: 'King Arthur',
+    quest: 'Find the Holy Grail',
+    favorite_color: 'Blue',
   }
-  writeLogEntry();
+
+  const json_Entry = log.entry(metadata, message);
+
+  // Synchronously write the log entry
+  await log.write(text_entry);
```
JustinBeckwith, line 54: As this is written right now, the user will see `await` calls without a wrapping async function.  This will lead to copy/paste code that doesn't work sadly.  This needs an outer `main` function that holds an async function that does an await.  I know this is kind of terrible.   For a good example:
https://github.com/googleapis/nodejs-bigquery/blob/master/samples/addColumnLoadAppend.js

nicoleczhu: Ah yes we want it to be copy&paste runnable.

@JustinBeckwith should I refactor all node tests to have an outer main (might take some time). 
Or is it ok if I wrap the await call inside an async function like the [current state of the example](https://cloud.google.com/logging/docs/samples/logging-write-log-entry#logging_write_log_entry-nodejs)? 

```js
...
async function writeLogEntry() {
  // Save the two log entries. You can write entries one at a time, but it is
  // best to write multiple entires together in a batch.
  await log.write([entry, secondEntry]);
  console.log(`Wrote to ${logName}`);
}
writeLogEntry();
``

File: samples/logs.js (hunk 2 above)
Comment (nicoleczhu) [reply samples/logs.js]: Note for future

File: samples/logs.js (hunk 1 above)
Comment (nicoleczhu) [reply samples/logs.js]: Ah, Github diff UI is a bit confusing. I didn't delete the placeholder. 

I just changed the logname copy from `'Name of the log to write to, e.g. my-log';` into `my-log` since the previous line also has a `TODO(developer)` comment that already gives context.

File: samples/logs.js (hunk 3 above)
Comment (JustinBeckwith) [reply samples/logs.js:54]: Shouldn't have to refactor the tests - just the samples themselves.  What you have written there is perfect.  That whole thing though ends up in it's own `main` function usually.

Comment (nicoleczhu) [reply samples/logs.js]: 👍  updated . thanks for the feedback!

**Review by JustinBeckwith — APPROVED** [review]

File: samples/logs.js [hunk 4]
```diff
@@ -16,44 +16,50 @@
...
+
+    // Synchronously batch write the log entries
+    await log.write([text_entry, json_Entry]);
+
+    // Asynchronously let the logging library dispatch logs
+    log.write(text_entry);
+
     console.log(`Wrote to ${logName}`);
   }
   writeLogEntry();
```
Daniel-Sanche, line 65: Is this supposed to be inside the region tag?

samples/logs.js
```diff
@@ -16,44 +16,50 @@
...
-  // Creates a client
   const logging = new Logging();
 
   /**
-   * TODO(developer): Uncomment the following line to run the code.
+   * TODO(developer): Uncomment the following line and replace with your values.
    */
-  // const logName = 'Name of the log to write to, e.g. my-log';
-
+  // const logName = 'my-log';
```
Daniel-Sanche, line 25: If you don't want to have this as a comment, this should also work:

```
async function writeLogEntry(inputName) {
...
var logName = 'my-log';
[END logging_write_log_entry]
logName = inputName;
[START logging_write_log_entry]
...
```

The two halves of the region tag will be concatenated together, and the snippet will be both readable and testable

samples/logs.js
```diff
@@ -16,44 +16,50 @@
...
-  // A structured log entry
-  const secondEntry = log.entry(
-    {resource: resource},
-    {
-      name: 'King Arthur',
-      quest: 'Find the Holy Grail',
-      favorite_color: 'Blue',
-    }
-  );
+  const json_Entry = log.entry(metadata, message);
```
Daniel-Sanche, line 51: More comments might be useful here. I wonder if users will be confused about the multiple logging payloads

samples/logs.js
```diff
@@ -16,44 +16,50 @@
...
+    // Synchronously write the log entry
+    await log.write(text_entry);
+
+    // Synchronously batch write the log entries
+    await log.write([text_entry, json_Entry]);
+
+    // Asynchronously let the logging library dispatch logs
+    log.write(text_entry);
+
     console.log(`Wrote to ${logName}`);
```
Daniel-Sanche, line 63: It seems a bit weird to be doing `log` and `console.log`. Is there any reason?

Issue #988, opened 2021-02-04
Support async iterators in handwritten layer for all "list x" calls


0xSage:
Create functions getEntriesAsync(), getLogsAsync() and getLogMetricAsync(), extending [listlogentriesasync](https://googleapis.dev/nodejs/logging/latest/v2.LoggingServiceV2Client.html#listLogEntriesAsync), listlogsasync, listlogmetricsasync, etc

Async iterators are available in the GAPIC portion of this lib. For example: users can now do by: 
```javascript
const {Logging, LoggingServiceV2} = require('@google-cloud/logging');
const logging = new Logging();

async function getlogsAsync() {
  const time = new Date();
  time.setHours(time.getHours() - 3); // change this up
  const timeFilter = `timestamp >= "${time.toISOString()}"`;

  const req = {
    resourceNames: ["projects/log-bench"],
    filter: timeFilter,
  }

  const iterable = logging.loggingService.listLogEntriesAsync(req);
  for await (const response of iterable) {
    console.log(response);
  }
}

getlogsAsync();
```

This is clunky. Write a similar wrapper for these funcs here: https://github.com/googleapis/nodejs-logging/blob/543533a02f293028081e7d31e3a98be1677b426a/src/index.ts#L590

losalex:
Closing this issue due to it is pretty stale and seems there is a lack of demand - feel free to reopen or comment if this feature is really required.

PR #998, opened 2021-02-18
feat: tail entries wrapper

### LINKED ISSUE #996: TailLogEntries wrapper
Write wrappers for logging and log for new api endpoint [TailLogEntries](https://cloud.google.com/logging/docs/reference/v2/rpc/google.logging.v2#google.logging.v2.LoggingServiceV2.TailLogEntries) 

- [ ] logging wrapper
- [ ] log wrapper
- [ ] handle suppressionInfo
- [ ] handle buffering
- [ ] handle filtering
- [ ] tests

freelerobot:
Fixes #996 🦕

Design doc: [go/cdpe-ops-spec-tailing-logs](https://docs.google.com/document/d/1xgbCjdZL7X18ihIhNzQvjwZs8Y1gG0V3wx130PhHtR4/edit?resourcekey=0-_dByx3OV_8ovx8cqgVnu9Q)

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -22,7 +22,7 @@ import arrify = require('arrify');
 import * as extend from 'extend';
 import * as gax from 'google-gax';
 // eslint-disable-next-line node/no-extraneous-import
-import {ClientReadableStream} from '@grpc/grpc-js';
+import {ClientReadableStream, ClientDuplexStream} from '@grpc/grpc-js';
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires
 const pumpify = require('pumpify');
@@ -44,6 +44,7 @@ import {Entry, LogEntry} from './entry';
 import {
   Log,
   GetEntriesRequest,
+  TailEntriesRequest,
   LogOptions,
   MonitoredResource,
   Severity,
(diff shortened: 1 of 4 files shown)


src/index.ts
```diff
@@ -22,7 +22,7 @@ import arrify = require('arrify');
 import * as extend from 'extend';
 import * as gax from 'google-gax';
 // eslint-disable-next-line node/no-extraneous-import
-import {ClientReadableStream} from '@grpc/grpc-js';
+import {ClientReadableStream, ClientDuplexStream} from '@grpc/grpc-js';
```
bcoe, line 25: curious why we need to use a duplex stream, when you're tailing do we also write to the stream?

src/index.ts
```diff
@@ -698,6 +704,104 @@ class Logging {
...
+        resourceNames: options.resourceNames,
+        ...(options.filter && {filter: options.filter}),
+        ...(options.bufferWindow && {bufferwindow: options.bufferWindow}),
+      };
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      if (!(global as any).GCLOUD_SANDBOX_ENV) {
+        gaxStream = this.loggingService.tailLogEntries(options.gaxOptions);
+        // Write can only be called once in a single tail streaming session.
+        gaxStream.write(writeOptions);
```
bcoe, line 796: that answers my question, if I'm following do you write options regarding how the stream will be tailed, and then listen on it for events?

system-test/logging.ts
```diff
@@ -386,6 +412,23 @@ describe('Logging', () => {
             done();
           });
       });
+
+      it('should tail log entries as a stream', done => {
```
bcoe, line 416: I'd be interested in being able to tail entries as an async iteterator:

```
for async (const entry of this.tailEntriesAsync()) {
  console.info(entry);
}
```

Could we add a test for this? I think it makes for a much nicer interface than streams -- for folks using libraries going forward.

bcoe (changes requested):
This looks great to me, this feature just didn't exist prior to this work?

src/index.ts
nicoleczhu: Correct. We have the option of writing options that define filters & bufferwindow for what/how to retrieve the logs.

system-test/logging.ts
nicoleczhu: sure sounds good!

nicoleczhu: Ah I think I already have an issue tracking it here : https://github.com/googleapis/nodejs-logging/issues/988

Pls cmiiw through, I dont think `tailLogEntries` has this the async iterator feature. It's only for `listX` which is a different API endpoint than this current one I'm implementing.

nicoleczhu: @bcoe can you let me know if I should make any other changes to this PR or give it a stamp? Thank you!

bcoe (approved):

bcoe:
@nicoleczhu apologies for the slow review, this looks good to me 👍

Issue #1003, opened 2021-03-16
Documentation needs lots of love, pleaaase. 🙏 


govindrai:
**Is your feature request related to a problem? Please describe.**
I'm pretty frustrated by _all_ the node.js client libraries' documentation. 

Let's take an example with the Node Logging Client. Let's say I want to create a sink. Searching "createSink" returns 5 different ways for me to create a sink: 
![image](https://user-images.githubusercontent.com/13859249/111334476-d4b33a80-8630-11eb-806b-ac2b482042fc.png)
![image](https://user-images.githubusercontent.com/13859249/111334541-ded53900-8630-11eb-8b3f-24a45e9ebdc3.png)
_This one has all methods doubly listed which may be a separate bug_
![image](https://user-images.githubusercontent.com/13859249/111334793-193ed600-8631-11eb-9f13-de135ab3a944.png)
![image](https://user-images.githubusercontent.com/13859249/111334850-252a9800-8631-11eb-9b16-720e2847aa8e.png)

Oh and you can also use Sink#create as well.
![image](https://user-images.githubusercontent.com/13859249/111334573-e85ea100-8630-11eb-8213-ecff4b4c42b6.png)

Can we cut down on the noise? There's nothing particularly complex that an end user is trying to do with these sdks. Just basic crud operations on resources. However, the docs are littered which 100s of methods to various classes that for the most part are not used directly (i.e. another method is calling them). 

The left hand side bar includes so many links to so many different classes, that to find any relevant apis, one has either scroll forever or read through a ton of classes they may never use. There's 122 classes listed in the sidebar.. For example when loading the docs for Logging, the first three class are:
![image](https://user-images.githubusercontent.com/13859249/111330226-fca09f00-862c-11eb-9872-ccde982ddc00.png)

You're most likely never going to interface with any of those classes directly (i.e there's another API method that instantiating and using those classes). Maybe it makes sense to link directly in the docs of those higher level API methods where they're relevant and not in the main catalogue. Or to list these other classes at the very bottom and have them be in a drawer that can be collapsed if needed. Or simply hide the internals.

Secondly there are so many classes (i.e. LoggingServiceV2Client, ConfigServiceV2Client, MetricServiceV2Client) and there's overlap in between them. What are these for? The examples don't use them, when are you supposed one over the other? Many of these classes don't have any examples of their usage. 

[...]

PR #1009, opened 2021-03-25
docs: add tailEntries sample and docs

### LINKED ISSUE #1000: Samples for tailEntries
Now that tailEntries is available, write samples and update docs for it

freelerobot:
Fixes #1000

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,3 +4,4 @@ test/fixtures
 build/
 docs/
 protos/
+env-tests-logging/
--- a/samples/logs.js
+++ b/samples/logs.js
@@ -152,6 +152,37 @@ async function listLogEntriesAdvanced(filter, pageSize, orderBy) {
   // [END logging_list_log_entries_advanced]
 }
 
+async function tailLogEntries(logName) {
+  // [START logging_tail_log_entries]
+  const {Logging} = require('@google-cloud/logging');
+  const logging = new Logging();
+
+  /**
+   * TODO(developer): Replace logName with the name of your log.
+   */
+  const log = logging.log(logName);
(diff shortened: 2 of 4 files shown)


samples/logs.js
```diff
@@ -152,6 +152,37 @@ async function listLogEntriesAdvanced(filter, pageSize, orderBy) {
   // [END logging_list_log_entries_advanced]
 }
 
+async function tailLogEntries(logName) {
```
bcoe, line 155: non-blocking suggestion, if you pull this into its own file, it will have its own listing in `README.md` and `samples/README.md` which can be nice.

.eslintignore
```diff
@@ -4,3 +4,4 @@ test/fixtures
 build/
 docs/
 protos/
+env-tests-logging/
```
bcoe, line 7: Our templates automatically populate this file, if you need this new directory you can add it to `excludes` an an extra argument [here](https://github.com/googleapis/nodejs-logging/blob/master/synth.py#L35).

bcoe (approved):
Left a couple non-blocking nits.

samples/logs.js
nicoleczhu: This is a great suggestion, thanks!

.eslintignore
nicoleczhu: updated

samples/logs.js
nicoleczhu: Actually, I think all of the samples need to be reorganized as a part of [this ticket](https://github.com/googleapis/nodejs-logging/issues/1003). I'll get the samples refactoring in in a separate PR if thats ok

PR #1021, opened 2021-04-01
fix(deps): drop dependency on is

JustinBeckwith:
I may be wrong, but I don't think we actually needed this

### DIFF
--- a/src/common.ts
+++ b/src/common.ts
@@ -14,8 +14,6 @@
  * limitations under the License.
  */
 
-import * as is from 'is';
-
 export interface ObjectToStructConverterConfig {
   removeCircular?: boolean;
   stringify?: boolean;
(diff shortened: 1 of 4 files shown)


src/common.ts
```diff
@@ -105,27 +103,35 @@ export class ObjectToStructConverter {
...
       };
-    } else if (is.object(value)) {
+    } else if (Array.isArray(value)) {
+      convertedValue = {
+        listValue: {
+          values: (value as Array<{}>).map(this.encodeValue_.bind(this)),
+        },
+      };
+    } else if (value?.toString() === '[object Object]') {
+      // Using `typeof value === 'object'` is discouraged here, because it
```
nicoleczhu, line 133: tricky!

freelerobot (approved):

src/common.ts
JustinBeckwith: Yeah, it's like I was saying - `typeof foo === 'object'` is a weird footgun, so I'm conflicted on the change.  This check more accurately matches what `is.object` does in other places.

PR #1023, opened 2021-04-01
fix(deps): remove dependency on through2

JustinBeckwith:
This drops the top level dependency on `through2`, though it's still technically brought in via `@google-cloud/common` and possibly `pumpify`.  It moves us towards standard Transform and PassThrough streams instead of the wrapper on top.  This is party of my "reduce all the dependencies" work. 

Note: this change is not without risk.  We've tried this before, and run into issues.  This library has solid system and sample tests - *it would be super helpful if someone could run those system tests on node 10, 12, and 14, just to do a confidence check*.


nicoleczhu:
Testing on node10 and 14 now. Automation already defaults to 12.

freelerobot (changes requested):
I tested on node10 and 14. Automation already defaults to testing on 12.
**Results**
- [x] Node10 system test works
- [x] Node10 sample test works
- [ ] Node14 system test works
- [x] Node14 sample test works

@JustinBeckwith FYI the following 1 test failed in Node14

```sh
1) Logging
       logs
         log-specific entries
           should list log entries as a stream:
     Error: Timeout of 600000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves. (/Users/nicolezhu/Desktop/nodejs-logging/build/system-test/logging.js)
      at listOnTimeout (internal/timers.js:554:17)
      at processTimers (internal/timers.js:497:7)
```

nicoleczhu:
FYI: I will resolve this as a part of my Q2 OKR to [reduce dependencies](https://docs.google.com/document/d/1TJXE7D8QGlh-y5BHLa4Yfq1x5-0rPt4pwUB1kekrLqQ/edit#) on node libraries

freelerobot (approved):

PR #1034, opened 2021-04-08
fix: accept uniqueWriterIdentity in setMetadata

freelerobot:
[Final Spec](https://docs.google.com/document/d/1Uzoa-7YTahpA7HwDY0P-CobJ6UccjSSqBqNFBdS3XCU/edit?usp=sharing) (internal googler link). Fixes: [b/183619158](https://buganizer.corp.google.com/issues/183619158)

Changes: 
- Users can provide uniqueWriterIdentity boolean in setmetadata
- Users can change sink into a uniqueWriterIdentity, from a shared identity

### DIFF
--- a/src/sink.ts
+++ b/src/sink.ts
@@ -34,6 +34,7 @@ export type SinkMetadataResponse = [LogSink];
 
 export interface SetSinkMetadata extends LogSink {
   gaxOptions?: CallOptions;
+  uniqueWriterIdentity?: boolean | string;
 }
 
 /**
@@ -289,6 +290,9 @@ class Sink {
   /**
    * Set the sink's metadata.
    *
+   * Note: If the sink was created with uniqueWriterIdentity = true, then the
+   * sink must be updated with uniqueWriterIdentity = true.
+   *
    * @see [Sink Resource]{@link https://cloud.google.com/logging/docs/reference/v2/rest/v2/projects.sinks#LogSink}
    * @see [projects.sink.update API Documentation]{@link https://cloud.google.com/logging/docs/reference/v2/rest/v2/projects.sinks/update}
    *
@@ -325,11 +329,18 @@ class Sink {
    */
   async setMetadata(metadata: SetSinkMetadata): Promise<SinkMetadataResponse> {
     const [currentMetadata] = await this.getMetadata();
-    const reqOpts = {
+    const uniqueWriterIdentity = metadata.uniqueWriterIdentity;
+    delete metadata.uniqueWriterIdentity;
+    let reqOpts = {
       sinkName: this.formattedName_,
       sink: extend({}, currentMetadata, metadata),
     };
     delete reqOpts.sink.gaxOptions;
+    // Add user specified uniqueWriterIdentity boolean, if any.
+    reqOpts = {
+      ...reqOpts,
+      ...(uniqueWriterIdentity && {uniqueWriterIdentity}),
+    };
     [this.metadata] = await this.logging.configService.updateSink(
       reqOpts,
       metadata.gaxOptions
--- a/src/sink.ts
+++ b/src/sink.ts
@@ -290,8 +290,10 @@ class Sink {
   /**
    * Set the sink's metadata.
    *
-   * Note: If the sink was created with uniqueWriterIdentity = true, then the
-   * sink must be updated with uniqueWriterIdentity = true.
+   * Note: If the sink was previously created or updated with
+   * uniqueWriterIdentity = true, then you must update the sink by setting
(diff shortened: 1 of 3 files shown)


src/sink.ts
```diff
@@ -325,11 +329,18 @@ class Sink {
...
+    let reqOpts = {
       sinkName: this.formattedName_,
       sink: extend({}, currentMetadata, metadata),
     };
     delete reqOpts.sink.gaxOptions;
+    // Add user specified uniqueWriterIdentity boolean, if any.
+    reqOpts = {
+      ...reqOpts,
+      ...(uniqueWriterIdentity && {uniqueWriterIdentity}),
+    };
```
stephenplusplus, line 343: We could just do:

```js
if (metadata.uniqueWriterIdentity) {
  reqOpts.uniqueWriterIdentity = metadata.uniqueWriterIdentity;
}
```

src/sink.ts
```diff
@@ -289,6 +290,9 @@ class Sink {
   /**
    * Set the sink's metadata.
    *
+   * Note: If the sink was created with uniqueWriterIdentity = true, then the
```
stephenplusplus, line 293: To me, this reads like the user will have to provide `uniqueWriterIdentity = true`, but it looks like the code does it for them. Should this say something like "Note: If the sink was created with `uniqueWriterIdentity = true`, we will automatically apply this option during the `setMetadata` operation."?

src/sink.ts
nicoleczhu: Doing that yields typescript error `TS2339: Property 'uniqueWriterIdentity' does not exist on type '{ sinkName: string; sink: ILogSink & SetSinkMetadata; }'.`

Would prefer to not have to formally define reqOpts since we're transiently using it in this function

src/sink.ts
nicoleczhu: We dont automatically apply this in the setMetadata operation yet. Hence [this issue](https://github.com/googleapis/nodejs-logging/issues/1035), but I need to double check with logging-eng whether my proposed detection heuristic is reliable. For the scope of this PR, I aim to unblock the user but letting them set the boolean manually.

Will update the comment to be clearer :)

src/sink.ts
```diff
@@ -34,6 +34,7 @@ export type SinkMetadataResponse = [LogSink];
 
 export interface SetSinkMetadata extends LogSink {
   gaxOptions?: CallOptions;
+  uniqueWriterIdentity?: boolean | string;
```
simonz130, line 37: add a comment to explain what this is

system-test/logging.ts
```diff
@@ -182,6 +182,37 @@ describe('Logging', () => {
...
+        const FILTER = 'severity = ALERT';
+        await sink.create({
+          destination: topic,
+        });
+        const metadata = {
+          filter: FILTER,
+          uniqueWriterIdentity: true,
+        };
+        const [apiResponse] = await sink.setMetadata(metadata);
+        assert.strictEqual(apiResponse.filter, FILTER);
```
simonz130, line 212: Not sure how you check that uniqueWriterIdentity was changed from false to true in this case. I see you compare filters and filter is string that doesn't include uniqueWriterIdentity

nicoleczhu: In this test, I only check that users are able to edit filter, given their uniqueWriterId settings. This was the bug that this PR fixes.

You've pinpointed the fundamental issue that uniqueWriterId isn't returned by any API calls at all (which was why I have [this issue](https://github.com/googleapis/nodejs-logging/issues/1035) to help users infer what that boolean was, still waiting on logging-eng to confirm if the inferrence heuristic is reliable)

src/sink.ts
nicoleczhu: oke!

nicoleczhu: Added

system-test/logging.ts
simonz130: perhaps, worth adding this explanation as comment to the bug or rephrase the name of the test.

system-test/logging.ts
```diff
@@ -182,6 +182,43 @@ describe('Logging', () => {
...
+        // Sink must be deleted within this test before any logs are generated
+        // to avoid topic_permission_denied emails.
+        await sink.delete();
```
simonz130, line 218: Can you have a finally block to always delete the sink at the end of the test?

simonz130 (approved):

system-test/logging.ts
nicoleczhu: I have an `after` block to delete sinks for the other tests. But these uniqueId sinks require a reference to the sink names  created within the scope of the tests.

Issue #1037, opened 2021-04-12
Reduce dependencies & library binary


nicoleczhu:
For this library to be a dependency for serverless (GCF/GCR) users, it needs to be smaller. Also we need to just reduce the number of dependencies in general. Many releases break semver and thus break the CI in this library.

nicoleczhu:
Internal doc on progress: go/cdpe-ops-node-q22021

PR #1038, opened 2021-04-14
chore(deps): remove snakecase-keys dependency

freelerobot:
Before: **Total dependencies size: 22.49M**
After: **Total dependencies size: 22.46M**

improvement: ~30kB

Part of #1037

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -26,9 +26,6 @@ import {Entry, EntryJson, LogEntry} from './entry';
 import {getDefaultResource} from './metadata';
 import {GoogleAuth} from 'google-auth-library/build/src/auth/googleauth';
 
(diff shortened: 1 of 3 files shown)


src/log.ts
```diff
@@ -928,6 +923,16 @@ class Log implements LogSeverityFunctions {
         options.gaxOptions
       );
     }
+    function snakecaseKeys(labels: {[p: string]: string} | null | undefined) {
+      for (const key in labels) {
```
simonz130, line 927: please document what is happening in this function

simonz130 (approved):

src/log.ts
nicoleczhu: added!

PR #1063, opened 2021-05-10
build: release minified package under new name

### LINKED ISSUE #1037: Reduce dependencies & library binary

freelerobot:
Fixes #1037  🦕

### DIFF
--- a/.kokoro/publish.sh
+++ b/.kokoro/publish.sh
@@ -28,4 +28,31 @@ NPM_TOKEN=$(cat $KOKORO_GFILE_DIR/secret_manager/npm_publish_token)
 echo "//wombat-dressing-room.appspot.com/:_authToken=${NPM_TOKEN}" > ~/.npmrc
 
 npm install
+
+# Publish the default package
+npm publish --access=public --registry=https://wombat-dressing-room.appspot.com
+
+# Uglify /build to publish a smaller package
+for d in $(find ./build/ -maxdepth 8 -type d)
+do
+  pushd $d
+    for f in *.js; do
+      [ -f "$f" ] || break
+
+      if [ -f "$f.map" ]; then
+        # Keep original .ts source mappings
+        uglifyjs --source-map "content='$f.map', url='$f.map'" "$f" --output "$f"
+      else
+        uglifyjs "$f" --output "$f"
+      fi
+    done
+  popd
+done
+
+# Change and publish under package name `@google-cloud/logging-min`
+sed -i -e 's/"name": "@google-cloud\/logging"/"name": "@google-cloud\/logging-min"/' package.json
+if [ -f "package.json-e" ]; then
+  rm package.json-e
+fi
+
 npm publish --access=public --registry=https://wombat-dressing-room.appspot.com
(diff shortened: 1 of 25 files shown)


simonz130 (approved):

.kokoro/publish.sh
```diff
@@ -28,4 +28,31 @@ NPM_TOKEN=$(cat $KOKORO_GFILE_DIR/secret_manager/npm_publish_token)
...
+        uglifyjs --source-map "content='$f.map', url='$f.map'" "$f" --output "$f"
+      else
+        uglifyjs "$f" --output "$f"
+      fi
+    done
+  popd
+done
+
+# Change and publish under package name `@google-cloud/logging-min`
+sed -i -e 's/"name": "@google-cloud\/logging"/"name": "@google-cloud\/logging-min"/' package.json
```
bcoe, line 53: This seems a little hacky to me, I was picturing we'd perhaps just make a new repository for this use case.

Is this the eventual plan?

.kokoro/publish.sh
```diff
@@ -28,4 +28,31 @@ NPM_TOKEN=$(cat $KOKORO_GFILE_DIR/secret_manager/npm_publish_token)
...
+# Uglify /build to publish a smaller package
+for d in $(find ./build/ -maxdepth 8 -type d)
+do
+  pushd $d
+    for f in *.js; do
+      [ -f "$f" ] || break
+
+      if [ -f "$f.map" ]; then
+        # Keep original .ts source mappings
+        uglifyjs --source-map "content='$f.map', url='$f.map'" "$f" --output "$f"
```
bcoe, line 44: I would prefer we pull the logic for minifying into `package.json` scripts, so that when you want the minified version you can just go:

`npm compile-min`

bcoe: This minifies the amount that the publication script deviates from our usual publication script.

bcoe (changes requested):
If we are going with this route, I'd prefer that the build logic is encapsulated in npm scripts.

My preference would be to eventually make this experiment its own GitHub repository, to give you more flexibility with regards to what a smaller version of the logging library looks like ... this feels gross.

_I can understand starting the experiment in the same repository, not blocking for me; but I'd rather we pull more logic into npm scripts, which will benefit you when/if we split this out into its own repository._

bcoe (approved):
Approving, so that I'm not blocking your work (_since I think we're still on different time zones._).

The actionable feedback I have is that I have a preference for us capturing the build process in npm scripts 👍

.kokoro/publish.sh
nicoleczhu: Good idea. My previous attempt did this - and we can just have a new publish-min script. Will modify

.kokoro/publish.sh
nicoleczhu: Yes this is the eventual plan. Doc shared for your review in our chat.

PR #1073, opened 2021-05-19
fix: debrand Stackdriver & do not throw errors

freelerobot:
dont throw errors in client libraries

### DIFF
--- a/src/http-request.ts
+++ b/src/http-request.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-export interface StackdriverHttpRequest {
+export interface CloudLoggingHttpRequest {
   requestMethod?: string;
   requestUrl?: string;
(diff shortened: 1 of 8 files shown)


src/http-request.ts
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-export interface StackdriverHttpRequest {
+export interface CloudLoggingHttpRequest {
```
daniel-sanche, line 17: Would this count as an API-breaking change?

nicoleczhu: I was wondering the same. It is not, as this type is only used internally and not re-exported to users by `index.ts`, and was never exposed in the [client docs here](https://googleapis.dev/nodejs/logging/latest/index.html).

simonz130 (approved):

PR #1086, opened 2021-06-03
feat: users can log raw http request objects with trace

### LINKED ISSUE #1083: feat: detect trace & span fields
Detect trace and span fields to populate log entries in the following situations, in increasing order of priority
1. User is using middleware that has access to request objs
2. User is directly logging an HttpRequest object
3. User explicitly declares trace & span in a custom log entry

freelerobot:
A. **Users can log raw http.incomingmessage (requests) now:**
1. Users can now log.write(entry.metadata.requestObj) where reqObj is a **raw incoming HTTP request** type. 
2. Users can now log.write(entry.metadata.requestObj) where reqObj is an **exported CloudLoggingHttpRequest** type.
3. The above mentioned request logs are auto-formatted into GCP compliant structured request logs. See LogEntry guidelines: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#httprequest
3. Lifted `make-http-requests` out of /middleware subdirectories. Since it’s not just for middleware. This is cleaner, as http-request was already in the main directory. 

B. **Users can force tracing by providing x-cloud-trace-context in http headers (in various formats):**
See: https://cloud.google.com/trace/docs/setup#force-trace 
1.  `trace`, `spanID` and `traceSampled` fields in X-Cloud-Trace-Context can be optional (see context/trace_context.proto)
2. TraceSample should default to false (verified with nwang@)
3. Trace/span IDs should still be included in log contexts even when traceSampled is false, because "A non-sampled trace value is still useful as a request correlation identifier." ([LogEntry spec](https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry))

Fixes: #1083

### DIFF
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -19,29 +19,40 @@ const EventId = require('eventid');
 import * as extend from 'extend';
 import {google} from '../protos/protos';
 import {objToStruct, structToObj} from './common';
+import {CloudLoggingHttpRequest} from './http-request';
+import {makeHttpRequestData, ServerRequest} from './make-http-request';
 import * as http from 'http';
 
 const eventId = new EventId();
 
+// Additional field types supported by this client library.
 export type Timestamp = google.protobuf.ITimestamp | Date | string;
 export type LogSeverity = google.logging.type.LogSeverity | string;
+export type HttpRequest =
+  | google.logging.type.IHttpRequest
+  | CloudLoggingHttpRequest
+  | ServerRequest;
+
 export type LogEntry = Omit<
   google.logging.v2.ILogEntry,
-  'timestamp' | 'severity'
+  'timestamp' | 'severity' | 'httpRequest'
 > & {
   timestamp?: Timestamp | null;
   severity?: LogSeverity | null;
+  httpRequest?: HttpRequest | null;
 };
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
 export type Data = any;
 
+// The final Entry format submitted to the LoggingService API.
 export interface EntryJson {
   timestamp: Timestamp;
   insertId: number;
   jsonPayload?: google.protobuf.IStruct;
   textPayload?: string;
-  trace?: string;
+  httpRequest?: google.protobuf.IStruct;
+  //  TODO: explicitly declare trace/span here as well
 }
 
 export interface ToJsonOptions {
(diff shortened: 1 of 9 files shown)


src/entry.ts
```diff
@@ -182,6 +200,45 @@ class Entry {
...
+        const regex = /([a-f\d]+)?(\/?([a-f\d]+))?(;?o=(\d))?/;
+        const match = rawReq.headers['x-cloud-trace-context']
+          .toString()
+          .match(regex);
+        if (match) {
+          if (!this.metadata.trace && match[1]) this.metadata.trace = match[1];
+          if (!this.metadata.spanId && match[3])
+            this.metadata.spanId = match[3];
+          if (this.metadata.traceSampled === undefined && match[5])
+            this.metadata.traceSampled = match[5] === '1';
```
simonz130, line 236: I don't think there is much sense in having traceSampled value as true or false in this.metadata.traceSampled but having trace and spanId not set.
Probably we can reduce the "if" to just 
```
if (match[5])
```

src/entry.ts
```diff
@@ -182,6 +200,45 @@ class Entry {
...
+      if (
+        'statusCode' in rawReq ||
+        'headers' in rawReq ||
+        'method' in rawReq ||
+        'url' in rawReq
+      )
+        this.metadata.httpRequest = makeHttpRequestData(rawReq);
+      // Infer trace & span if not user specified already.
+      if ('headers' in rawReq && rawReq.headers['x-cloud-trace-context']) {
+        const regex = /([a-f\d]+)?(\/?([a-f\d]+))?(;?o=(\d))?/;
```
simonz130, line 227: I suggest you refactor this to prepare for recognition of trace/span from traceparent (the recognition itself might happen in a different PR).

simonz130 (approved):
Thank you Nicole!

src/entry.ts
nicoleczhu: the users can still submit a header of ';o=0'. Which is a redundant but valid way to indicate both i) there's no trace/span and ii) don't sample this trace and span

src/entry.ts
nicoleczhu: ah ok, ill do this in a following pr with a TODO note on recognizing `trace-parent`

PR #1088, opened 2021-06-07
feat: detect Google and W3C trace context (including in middleware)

### LINKED ISSUE #1090: Change makeMiddleware api surface to include span and traceSampled

### LINKED ISSUE #1097: Make traceSampled injection logic more efficient
In the middleware scenario, we **auto inject trace/spans when users don't provide their own** `traceparent` or `x-cloud-trace-context`. See `make-middleware-express.ts`.

Every time we have to inject trace context, we make it sampled. This reveals a larger problem we need to think about (later) - on how to manage trace sampling logic in logging payloads.

The problem: at scale if billions of log entries will be generated with new trace context that is sampled - Cloud Trace will receive too many similar traces that might cause customers to pay for volume with no meaningful insights on service behavior.

In the meantime, users can manually configure `traceSampled` to `false` on their own.

freelerobot:
Fixes: #1090, #1097

Changes: 
- extends all span/trace detection for middleware scenario
- detects `traceparent` per W3C protocol with compatible extraction for this library
- detect `x-cloud-trace-context` per google protocol with compatible extraction for this library
- option to inject a generated `x-cloud-trace-context` into logentry
- drops @opencensus/propagation-stackdriver  dependency


src/entry.ts
```diff
@@ -214,15 +214,16 @@ class Entry {
...
       if (
         'statusCode' in rawReq ||
         'headers' in rawReq ||
         'method' in rawReq ||
         'url' in rawReq
       )
         this.metadata.httpRequest = makeHttpRequestData(rawReq);
-      // Infer trace & span if not user specified already.
+      // Infer trace & span if not user specified already
+      // TODO(nicolezhu): also infer from trace header if OpenCensus is configured
```
simonz130, line 226: can you open an issue on this in the repo?

src/middleware/context.ts
```diff
@@ -14,6 +14,17 @@
...
  */
 
+/*
+  When users are using middleware and this client library has access to request
+  headers, we should automatically extract as much information from the headers
+  as possible. This is in addition to when the user explicitly logs an
+  HTTPRequest logEntry object.
+
+  Scenarios:
+  - Context is available from Opencensus
```
simonz130, line 24: OpenCensus or OpenTelemetry (or in other words, from W3 Context propagation protocol)

src/middleware/express/make-middleware.ts
```diff
@@ -54,16 +58,22 @@ export function makeMiddleware<LoggerType>(
 
     const spanContext = getOrInjectContext(wrapper);
     const trace = `projects/${projectId}/traces/${spanContext.traceId}`;
+    const span = spanContext.spanId;
```
simonz130, line 61: can we extract isSampled field from the context? If the incoming request contains that, we should propagate its value downstream along with trace and span.

src/entry.ts
nicoleczhu: opened!

src/middleware/context.ts
nicoleczhu: added the comment

src/middleware/express/make-middleware.ts
nicoleczhu: yeah I think we should. will add it

src/http-request.ts
```diff
@@ -31,3 +45,194 @@ export interface CloudLoggingHttpRequest {
...
+export function getOrInjectTraceParent(
+  headerWrapper: HeaderWrapper,
+  inject?: boolean
+): w3cContext.SpanContext | null {
+  let spanContext = w3cContext.extract(headerWrapper);
+  if (spanContext) return spanContext;
+  if (inject) {
+    // We were the first actor to detect lack of context. Establish context.
+    spanContext = w3cContext.generate();
+    w3cContext.inject(headerWrapper, spanContext);
```
simonz130, line 235: I like it but behavior might be not consistent:
If **x-cloud-trace-context** is present, we extract it and pass over.
If not present, with this logic, we'll create a new traceparent (different format for context propagation than what Cloud Trace supports natively).

So what can happen:
- A system has service A that gets incoming requests from external systems with trace info specified in Cloud Trace format. In addition, there is periodic job that triggers the middleware w/o providing the trace/span info.
- When the request comes with x-cloud-trace-context, it will be propagated downstream in format X
- When periodic job is triggered, the request has no trace/span, so it will be propagated with traceparent - a format Y != X. 

We probably need to make it configurable - the default trace format to use when creating a new context.

test/http-request.ts
```diff
@@ -0,0 +1,332 @@
...
+      const cloudReq = makeHttpRequestData(req);
+      assert.strictEqual(cloudReq.protocol, 'http:');
+      assert.strictEqual(cloudReq.requestUrl, 'http://google.com/');
+      assert.strictEqual(cloudReq.requestMethod, 'GET');
+      assert.strictEqual(cloudReq.userAgent, undefined);
+      assert.strictEqual(cloudReq.referer, undefined);
+      assert.strictEqual(cloudReq.status, undefined);
+    });
+    it('should infer as many response values as possible', () => {
+      const RESPONSE_SIZE = 2048;
```
simonz130, line 78: size in what units?

test/http-request.ts
```diff
@@ -0,0 +1,332 @@
...
+  };
+
+  const fakeContext = Object.assign({}, FAKE_CONTEXT);
+
+  const {getOrInjectTraceParent} = proxyquire('../src/http-request', {
+    '@opencensus/propagation-stackdriver': fakeContext,
+  });
+
+  describe('getTraceContext', () => {
+    it('should return an empty trace string when extraction fails and inject is false', () => {
```
simonz130, line 138: can you rephrase to be more clear? which extraction?

test/http-request.ts
```diff
@@ -0,0 +1,332 @@
...
+    it('should return null if header property is not in http request', () => {
+      const req = {
+        method: 'GET',
+      } as http.IncomingMessage;
+      const wrapper = makeHeaderWrapper(req as unknown as http.IncomingMessage);
+      assert.strictEqual(wrapper, null);
+    });
+  });
+
+  describe('getCloudTraceContext', () => {
```
simonz130, line 195: Perhaps, worth being more descriptive here

test/http-request.ts
```diff
@@ -0,0 +1,332 @@
...
+            `From ${test.header}; Expected traceSampled: ${test.expected.traceSampled}; Got: ${context.traceSampled}`
+          );
+        } else {
+          assert(false);
+        }
+      }
+    });
+  });
+
+  describe('getOrInjectTraceParent', () => {
```
simonz130, line 285: similarly, please be more descriptive about what is being tested in this series of tests? Method name is not enough - we should describe the use-case (e.g. w3 trace extraction and propagation)

src/http-request.ts
```diff
@@ -31,3 +45,194 @@ export interface CloudLoggingHttpRequest {
...
+  if (wrapper) {
+    // Detect 'X-Cloud-Trace-Context' header.
+    const cloudContext = getCloudTraceContext(wrapper);
+    if (cloudContext) {
+      context.trace = `projects/${projectId}/traces/${cloudContext.trace}`;
+      context.spanId = cloudContext.spanId;
+      context.traceSampled = cloudContext.traceSampled;
+    } else {
+      // Detect 'traceparent' header, optionally injecting one if not available.
+      const parentContext = getOrInjectTraceParent(wrapper, inject);
```
simonz130, line 180: Perhaps, worth extracting the entire trace/span detection and propagation part into a separate method. It will also make it easier to test.

src/http-request.ts
```diff
@@ -31,3 +45,194 @@ export interface CloudLoggingHttpRequest {
...
+      req.headers[name] = value;
+    },
+    getHeader(name: string) {
+      return req.headers[name];
+    },
+  };
+  return wrapper;
+}
+
+export interface CloudTraceContext {
```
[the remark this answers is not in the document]
nicoleczhu: added in HEAD

src/http-request.ts
```diff
@@ -31,3 +45,194 @@ export interface CloudLoggingHttpRequest {
...
+      spanId: match[3],
+      traceSampled: match[5] === '1',
+    };
+  } else {
+    return null;
+  }
+}
+
+/**
+ * getCloudTraceContext looks for the HTTP header 'traceparent'
```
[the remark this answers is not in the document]
nicoleczhu: added in HEAD

test/http-request.ts
nicoleczhu: it just mocks the content length header, e.g. `content-length: 6553`, there's no unit that needs to be specified.

test/http-request.ts
nicoleczhu: the describe block should reflect the name of the function we're testing. I'll add some descriptive comments within the block

test/http-request.ts
nicoleczhu: It's convention to describe `function name` and then be more descriptive within the it() statements. I'm going to pepper more comments throughout the tests, but am not keen to change the describe snippets. 

It's impt to keep the current readability: 

Screen Shot 2021-06-09 at 10 45 53 PM

src/http-request.ts
nicoleczhu: that's a good suggestion! will do

test/http-request.ts
nicoleczhu: rephrased!

src/http-request.ts
nicoleczhu: Right now, we have to automatically inject `traceparent` if user doesn't have trace context && is using middleware. This is the default/previous behavior so turning it off would be a breaking change. 

In the non-middleware scenario, we actually don't inject a w3c context into the header (see `entry.ts`. The inject parameter is an optional bool, see:
```js
export function getOrInjectTraceParent(
  headerWrapper: HeaderWrapper,
  inject?: boolean
): w3cContext.SpanContext | null {
```

src/http-request.ts
```diff
@@ -31,3 +45,194 @@ export interface CloudLoggingHttpRequest {
...
+  trace: string;
+  spanId?: string;
+  traceSampled?: boolean;
+}
+
+/**
+ * getTraceContext returns a CloudTraceContext with as many available trace and
+ * span properties as possible. It examines HTTP headers for trace context.
+ * Optionally, it can inject a W3C compliant trace context when no context is
+ * available from headers.
```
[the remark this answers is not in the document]
nicoleczhu: Then we don't inject and we return the trace context detected. See [this line](https://github.com/googleapis/nodejs-logging/pull/1088/files#diff-0cd594dddd6a9f2e3e26afebbb92716434437d7001c2416252aa7531e7f2b8d4R143)

src/http-request.ts
```diff
@@ -14,6 +14,20 @@
...
+ * HTTP headers.
+ *
+ * Example:
+ * - If header context is available from OpenCensus or OpenTelemetry (via W3
+ * context propagation protocol), extract a log trace field.
+ * - If context is available from X-Cloud-Trace-Context header, extract trace,
+ * spanId, and traceSampled fields.
+ */
+
+import * as w3cContext from '@opencensus/propagation-stackdriver';
```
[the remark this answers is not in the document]
nicoleczhu: 1. OK, I rewrote `context.ts` to deprecate @opencensus/propagation-stackdriver . We depended on it previously, so this my impl continues supporting those features, but also does a lot more, e.g. it parses more `x-cloud-trace-context` formats, etc.

2. Also I spent some time looking into which otel packages we can use, incl otel - `api/core/tracing`. BUT, it’s hacky to directly use the new OTel node library to get and inject `traceparents`. They have hardcoded constructs that don’t map to ours. e.g. layers of abstraction between the header and the final trace properties [see this](https://github.com/open-telemetry/opentelemetry-js/blob/8e5c8a0965c2dc7517dd9f2084aebe56f678be89/packages/opentelemetry-core/src/trace/HttpTraceContextPropagator.ts#L73)

3. So I've extended the regex logic in this own library. Which gives us the added benefit of dropping the (large) opencensus dependency. 

4. Lastly, I realized we should actually be defaulting to injecting the `x-cloud-trace-context` when no trace context is avail in the middleware scenario. Not `traceparent`. This was the behavior before.

simonz130 (approved):

src/context.ts
```diff
@@ -0,0 +1,236 @@
...
+  const wrapper = makeHeaderWrapper(req);
+  if (wrapper) {
+    // Detect 'traceparent' header.
+    const traceContext = getContextFromTraceParent(wrapper, projectId);
+    if (traceContext) return traceContext;
+    // Detect 'X-Cloud-Trace-Context' header.
+    const cloudContext = getContextFromXCloudTrace(wrapper, projectId);
+    if (cloudContext) return cloudContext;
+    // Optional: Generate and inject a context for the user;
+    if (inject) {
```
daniel-sanche, line 110: It looks like inject is set explicitly by a bool flag. Should it then be the prioritized option, since users explicitly asked for it?

src/http-request.ts
```diff
@@ -31,3 +40,75 @@ export interface CloudLoggingHttpRequest {
...
+  // OriginalURL overwrites inferred url
+  if ('originalUrl' in req && req.originalUrl) {
+    requestUrl = req.originalUrl;
+    const url = new URL(requestUrl);
+    protocol = url.protocol;
+  }
+  req.method ? (requestMethod = req.method) : null;
+  if (req.headers && req.headers['user-agent']) {
+    req.headers['user-agent'] ? (userAgent = req.headers['user-agent']) : null;
+    req.headers['referer'] ? (referer = req.headers['referer']) : null;
```
daniel-sanche, line 87: nit: it might be good to make constants for these header strings

daniel-sanche:
Some of these behaviors seem like the kind of thing we'd want to be consistent across languages. It might be useful to have environment tests for these. But we can probably address that later

src/context.ts
```diff
@@ -0,0 +1,236 @@
...
+ * - In the middleware scenario, we auto-inject a Google trace context if one
+ * is not available.
+ */
+
+import * as http from 'http';
+import * as uuid from 'uuid';
+import * as crypto from 'crypto';
+
+/** Header that carries span context across Google infrastructure. */
+export const X_CLOUD_HEADER = 'x-cloud-trace-context';
```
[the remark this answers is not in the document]
nicoleczhu: oke!

src/context.ts
```diff
@@ -0,0 +1,236 @@
...
+  return context;
+}
+
+/**
+ * makeCloudTraceHeader generates valid X-Cloud-Trace-Context headers
+ */
+function makeCloudTraceHeader(): string {
+  const trace = uuid.v4().replace(/-/g, '');
+  const spanId = spanRandomBuffer().toString('hex');
+  return `${trace}/${spanId};o=1`;
```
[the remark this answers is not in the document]
nicoleczhu: opening

src/context.ts
nicoleczhu: this function is invoked internally only and not exported to users. but good pt - ill make it clearer in comments that it is only meant to kick in as a last option

PR #1106, opened 2021-06-24
docs: mention LogSync in readme

freelerobot:

### DIFF
--- a/.readme-partials.yml
+++ b/.readme-partials.yml
@@ -25,3 +25,29 @@ body: |-
 
   The `@google-cloud/logging` library will handle batching and dispatching
   these log lines to the API.
+
+  ## Writing to Stdout
+
+  Serverless applications like Cloud Functions, Cloud Run, and App Engine
+  should use the `LogSync` class to synchronously log to `stdout` or another
+  custom transport. Logs written to `stdout` are picked up by a [Logging agent](https://cloud.google.com/logging/docs/agent/logging)
+  injected with additional context, and streamed to the Logging API.
+
+  This library's `LogSync` gives users idiomatic syntax sugar, an on/off toggle
+  between writing to the API vs to `stdout`, and extracts additional log
+  properties like trace context from HTTP headers.
+
+  Read more about [structured logging](https://cloud.google.com/logging/docs/structured-logging).
+
+  ```js
+  // Optional: Create and configure a client
+  const logging = new Logging();
+  await logging.setProjectId()
+  await logging.setDetectedResource()
+
+  // Create a LogSync transport, defaulting to `process.stdout`
+  const log = logging.logSync(logname);
+  const meta = { // optional field overrides here };
+  const entry = log.entry(meta, 'Your log message');
+  log.write(entry);
+  ```
--- a/.readme-partials.yml
+++ b/.readme-partials.yml
@@ -28,14 +28,20 @@ body: |-
 
   ## Writing to Stdout
 
-  Serverless applications like Cloud Functions, Cloud Run, and App Engine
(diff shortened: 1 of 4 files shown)


.readme-partials.yml
```diff
@@ -25,3 +25,29 @@ body: |-
 
   The `@google-cloud/logging` library will handle batching and dispatching
   these log lines to the API.
+
+  ## Writing to Stdout
+
+  Serverless applications like Cloud Functions, Cloud Run, and App Engine
+  should use the `LogSync` class to synchronously log to `stdout` or another
+  custom transport. Logs written to `stdout` are picked up by a [Logging agent](https://cloud.google.com/logging/docs/agent/logging)
```
simonz130, line 33: The logging agent you refer to here is not the logging agent that is picking up the logs in serverless scenarios. Can you rephrase this to mention that it's picked out-of-process by a logging agent in respective serverless environment?

.readme-partials.yml
```diff
@@ -25,3 +25,29 @@ body: |-
...
+  between writing to the API vs to `stdout`, and extracts additional log
+  properties like trace context from HTTP headers.
+
+  Read more about [structured logging](https://cloud.google.com/logging/docs/structured-logging).
+
+  ```js
+  // Optional: Create and configure a client
+  const logging = new Logging();
+  await logging.setProjectId()
+  await logging.setDetectedResource()
```
simonz130, line 46: if someone doesn't do it, will this be calculated every time a call to log.write is happening?

.readme-partials.yml
```diff
@@ -25,3 +25,29 @@ body: |-
...
+  ## Writing to Stdout
+
+  Serverless applications like Cloud Functions, Cloud Run, and App Engine
+  should use the `LogSync` class to synchronously log to `stdout` or another
+  custom transport. Logs written to `stdout` are picked up by a [Logging agent](https://cloud.google.com/logging/docs/agent/logging)
+  injected with additional context, and streamed to the Logging API.
+
+  This library's `LogSync` gives users idiomatic syntax sugar, an on/off toggle
+  between writing to the API vs to `stdout`, and extracts additional log
+  properties like trace context from HTTP headers.
```
simonz130, line 38: does it provide the methods for logging at different verbosity levels? e.g. Log.info, Log.warning, etc?

.readme-partials.yml
```diff
@@ -25,3 +25,29 @@ body: |-
 
   The `@google-cloud/logging` library will handle batching and dispatching
   these log lines to the API.
+
+  ## Writing to Stdout
+
+  Serverless applications like Cloud Functions, Cloud Run, and App Engine
+  should use the `LogSync` class to synchronously log to `stdout` or another
```
simonz130, line 32: While this is true for serverless, this is also true for GKE, GCE and other platforms. By default we want folks to use stdout logging and let the agent pick it up - unless a particular scenario requires sending logs directly to Cloud Logging (e.g. hybrid, transitions between one logging vendor to another, etc) Can you rephrase this a bit?

simonz130 (approved):

.readme-partials.yml
nicoleczhu: yeah good pt, lemme say 'or'

.readme-partials.yml
nicoleczhu: Yes in the async / api log class `Log`, these calls are memoized

But in LogSync, these are of course optional

.readme-partials.yml
nicoleczhu: yes it does! lemme add that

.readme-partials.yml
nicoleczhu: ok!

PR #1119, opened 2021-07-14
fix: missing "uuid" dependency

koistya:
Fixes missing `uuid` dependency used in: https://github.com/googleapis/nodejs-logging/blob/5a0663aafd25d7b09307d94e1efe99118d71e528/src/utils/context.ts#L30-L32


bcoe (approved):

bcoe:
@koistya thank you for catching this 👏

PR #1145, opened 2021-09-01
fix: Correct detection of Cloud Run environment

### LINKED ISSUE #1127: `logging.detectedResource` does not return "cloud_run_revision" on Cloud Run
#### Environment details

  - OS: Container (base node:12-slim)
  - Node.js version: 12
  - npm version: 7.18.1
  - `@google-cloud/logging` version: 9.5.4

#### Steps to reproduce

  1. Create service w/ code snippet (see full service [here](https://github.com/averikitsch/nodejs-docs-samples/blob/cloud-logging/run/helloworld/index.js)):
```
  const {Logging} = require('@google-cloud/logging');
  await logging.setProjectId();
  await logging.setDetectedResource();
  console.log(logging.projectId);
  console.log(logging.detectedResource); // { type: 'global' }
```
  2. Deploy to Cloud Run:
```
gcloud beta run deploy logging-test --source .
```

3. Inspect logs:
```
console.log(logging.detectedResource); 
// outputs { type: 'global' }
```

minherz:
Uses dedicated GCPEnv enumerator (CLOUD_RUN).
Removes code that used COMPUTE_ENGINE enum with extra tests.
Updates relevant unit tests.

Fixes #1127

### DIFF
--- a/src/utils/metadata.ts
+++ b/src/utils/metadata.ts
@@ -193,15 +193,10 @@ export async function getDefaultResource(auth: GoogleAuth) {
       return getGAEDescriptor().catch(() => getGlobalDescriptor());
     case GCPEnv.CLOUD_FUNCTIONS:
       return getCloudFunctionDescriptor().catch(() => getGlobalDescriptor());
+    case GCPEnv.CLOUD_RUN:
+      return getCloudRunDescriptor().catch(() => getGlobalDescriptor());
     case GCPEnv.COMPUTE_ENGINE:
-      //  Note: GCPEnv.COMPUTE_ENGINE returns `true` for Google Cloud Run
-      if (process.env.K_CONFIGURATION) {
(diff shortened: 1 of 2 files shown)


daniel-sanche (commented):
I tested this manually and it seems to work!

There is currently a gap in the environment tests that should have caught this bug though, so I'm a bit hesitant to merge until that's fixed. If it's alright with you, I think we should hold off merging this one until I can fix up the tests, so we can see them go from red to green.

minherz:
@daniel-sanche which env test are you referencing. it looks like there are no such tests. in general, i prefer fast merges to avoid unnecessary maintenance. if you can point me to the right example, i can look into adding it.
what is a story with ci/kokoro tests?

daniel-sanche (approved):
I enabled environment tests as a presubmit check, and it looks like your changes took the cloud run test from [red](https://github.com/googleapis/nodejs-logging/pull/1148) to green. Thanks!

PR #1177, opened 2021-11-04
fix: Setting maxEntrySize does not truncate big json payloads correctly

### LINKED ISSUE #1087: Setting maxEntrySize does not truncate big json payloads correctly
When setting maxEntrySize for a logger and using `jsonPayload`, the truncation algorithm does not truncate all fields, but rather only a very small set of fields: https://github.com/googleapis/nodejs-logging/blob/master/src/log.ts#L984

As a user, when passing maxEntrySize, I expect it to also truncate custom fields. In our case, one of our fields is a huge list of UUIDs that does not matter if it gets truncated a little, but should be there. However, it does not get truncated and in the end we get exceptions thrown in prod because our log entries are too big.

losalex:
Add an ability to provide customized list of entries to be truncated. The list will be prepended to existing list of entries we truncate today by default, thus custom-defined fields will be truncated first

Fixes [#1087](https://github.com/googleapis/nodejs-logging/issues/1087) 🦕


src/log.ts
```diff
@@ -112,6 +116,31 @@ class Log implements LogSeverityFunctions {
...
+      'jsonPayload.fields.err.structValue.fields.stack.stringValue',
+      'jsonPayload.fields.err.structValue.fields.message.stringValue',
+      // All:
+      'jsonPayload.fields.message.stringValue',
+    ];
+
+    // Prepend all custom fields to be truncated to a list with defaults, thus
+    // custom fields will be truncated first. Make sure to filter out fields
+    // which are not in EntryJson.jsonPayload
+    if (options.jsonFieldsToTruncate !== undefined) {
```
minherz, line 133: what supposed to happen if `jsonFieldsToTruncate` is passed as `null`?

src/log.ts
```diff
@@ -988,17 +1017,7 @@ class Log implements LogSeverityFunctions {
...
-          'jsonPayload.fields.msg.stringValue',
-          'jsonPayload.fields.err.structValue.fields.stack.stringValue',
-          'jsonPayload.fields.err.structValue.fields.message.stringValue',
-          // All:
-          'jsonPayload.fields.message.stringValue',
-        ];
-        for (const field of fieldsToTruncate) {
+        for (const field of this.jsonFieldsToTruncate) {
           const msg: string = dotProp.get(entry, field, '');
           if (msg !== '') {
```
minherz, line 1022: nit: i think that `null` will pass this test together with non-empty strings

test/log.ts
```diff
@@ -633,6 +646,41 @@ describe('Log', () => {
       assert.ok(message.startsWith('hello world'));
       assert.ok(message.length < maxSize + entryMetaMaxLength);
     });
+
+    it('should not contin duplicate or illegal fields to be truncated and defaults should present', async () => {
+      assert.ok(log.jsonFieldsToTruncate.length > 1);
```
minherz, line 651: nit: would it be possible to assert the exact length instead? it should be a length of the predefined + 1, right?

minherz (changes requested):
please check for conditions that allow `null`.

src/log.ts
losalex: Yep, fixed - thanks!

src/log.ts
losalex: Perhaps I did not understood your comment, but `this.jsonFieldsToTruncate` should not be null... It can be directly accessed and set to null, but I guess this is not something we should handle.
I added a check for null strings when we update `this.jsonFieldsToTruncate`, let me know if thats what you mean to cover.

test/log.ts
losalex: We can - the only inconvenience will be broken tests every time we add values to `log.jsonFieldsToTruncate`... 
It would make sense to check exact amount of fields if by adding extra field we would have some extra steps to perform (e.g. adding some handler per entry), but in this case adding an entry is kinda no brainer.

minherz: yes, this could serve a kind of control to make people review the test if they make such change. leave it to your considerations.

src/log.ts
minherz: Sorry for misleading you. I meant the line 1022: `if (msg !== '') {`

minherz (approved):

src/log.ts
```diff
@@ -988,17 +1021,7 @@ class Log implements LogSeverityFunctions {
...
```
minherz, line 1026: This is the validation that allows `null`.

minherz (approved):
Please address a `null` validation in line 1022 (see the discussion in [review comments](https://github.com/googleapis/nodejs-logging/pull/1177#discussion_r774749486))

test/log.ts
losalex: Thanks - I will leave it as is for now, but will keep it in mind if we will ever need this as a way to control additions

src/log.ts
losalex: Sure, fixing

PR #1187, opened 2021-12-01
fix 186981671: The use of "synchronously" is a bit wrong on the help page

losalex:
This is a fix for wrong comment on https://cloud.google.com/logging/docs/reference/libraries

### DIFF
--- a/samples/logs.js
+++ b/samples/logs.js
@@ -51,10 +51,10 @@ async function writeLogEntry(logName) {
   const json_Entry = log.entry(metadata, message);
 
   async function writeLogEntry() {
-    // Synchronously write the log entry
+    // Asynchronously write the log entry
     await log.write(text_entry);
 
-    // Synchronously batch write the log entries
+    // Asynchronously batch write the log entries
     await log.write([text_entry, json_Entry]);
 
     // Asynchronously let the logging library dispatch logs


daniel-sanche (approved):

samples/logs.js
```diff
@@ -51,10 +51,10 @@ async function writeLogEntry(logName) {
   const json_Entry = log.entry(metadata, message);
 
   async function writeLogEntry() {
-    // Synchronously write the log entry
+    // Asynchronously write the log entry
```
minherz, line 54: I think that using "Asynchronously" here is incorrect as well. The following line
https://github.com/googleapis/nodejs-logging/blob/7f83b2bebade7879771011083cd30591034cdbfa/samples/logs.js#L55
does not perform asynchronously but blocks until the asynchronous operation completes (successfully or not).
The suggested (in the bug) "Write log entry and wait..." sounds better from technical perspective. For example, "Write log entry and wait for async writing to complete".

See also the following code with the similar comment but performing differently
https://github.com/googleapis/nodejs-logging/blob/7f83b2bebade7879771011083cd30591034cdbfa/samples/logs.js#L60-L61

minherz (commented):
please, DO NOT merge YET

samples/logs.js
losalex: The function call does performs asynchronously and not supposed to block execution of the thread, which means that the flow will be suspended and resume only once async operation is complete (this is the whole idea of async/await pattern)... Also I believe the statement "Write log entry and wait for async writing to complete" should be derived by programer from "await log.write(text_entry);" statement if developer familiar with async programming.
The original issue was reported on incorrect wording by using "Synchronously" which is fixed.
As for `log.write` without await, it can be addressed in separate PR.

PR #1189, opened 2021-12-02
fix: Correct a comment in sample regarding asynchronous log writing

losalex:
Correct another comment in sample regarding async write which does not awaits for operation to complete. Original issue surfaced in https://cloud.google.com/logging/docs/reference/libraries

### DIFF
--- a/samples/logs.js
+++ b/samples/logs.js
@@ -57,7 +57,7 @@ async function writeLogEntry(logName) {
     // Asynchronously batch write the log entries
     await log.write([text_entry, json_Entry]);
 
-    // Let the logging library dispatch logs synchronously
+    // Let the logging library dispatch logs
     log.write(text_entry);
 
     console.log(`Wrote to ${logName}`);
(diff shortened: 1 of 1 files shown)


samples/logs.js
```diff
@@ -57,7 +57,7 @@ async function writeLogEntry(logName) {
     // Asynchronously batch write the log entries
     await log.write([text_entry, json_Entry]);
 
-    // Asynchronously let the logging library dispatch logs
+    // Let the logging library dispatch logs synchronously
```
losalex: Based on fact that async/await pattern is Promise based, when `await` is not used with an async function call the function will execute synchronously (see some explanation [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function)). If function itself uses async calls with `await` keyword, function will execute asynchronously (see more insights [here](https://developer.mozilla.org/en-US/docs/Learn/JavaScript/Asynchronous/Async_await))

minherz (approved):
lgtm

PR #1191, opened 2021-12-13
chore: add additional option to populate http request metadata

### LINKED ISSUE #1129: HTTP logging and log-correlation needs better documentation
**Is your feature request related to a problem? Please describe.**
The [HTTP request sample](https://github.com/googleapis/nodejs-logging/blob/master/samples/http-request.js) shows constructing your own metadata object. This is not a great developer experience since many logging libraries can automatically do that for you. However, found via the [library tests](​​https://github.com/googleapis/nodejs-logging/blob/ad1067bf5820c41b3a80546feb259c76607fa338/test/entry.ts#L294), you can pass in the `req` object: `entry.metadata.httpRequest = req;`. 

 **Describe the solution you'd like**
This solution should be added or mentioned in the sample. The READMe or docs should also mention that this solution also does log-request correlation automatically for you.

minherz:
Adds clarification about more than one way to populate Http request metadata of the log entry.

Fixes #1129

### DIFF
--- a/.readme-partials.yml
+++ b/.readme-partials.yml
@@ -61,3 +61,13 @@ body: |-
   log.alert(entry);
   log.warning(entry);
   ```
+
+  ## Populating Http request metadata
+
+  Metadata about Http request is a part of the [structured log info](https://cloud.google.com/logging/docs/structured-logging)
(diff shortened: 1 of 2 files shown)


losalex (approved):
LGTM

Issue #1207, opened 2022-01-10
logSync ignores log_id when it generates logName field


anton-yarchuk:
`logging.logSync` function ignores provided `log_name` and generates standard `logName` field.

#### Environment details

  - OS: Ubuntu 20.4
  - Node.js version: 12.22.5
  - npm version: 6.14.14
  - `@google-cloud/logging` version: 9.6.6

#### Steps to reproduce

```js
const logging = new Logging();

await logging.setProjectId(); // I've tested with and without these two lines
await logging.setDetectedResource();

const log = logging.logSync('hello-log-sync');

const meta = {
  severity: 'INFO',
  labels: {
    iHaveLabels: 'too',
  },
};
const data = {
  some: 'data',
};

await log.write(log.entry(meta, data));
  ```

#### Expected results:

- 'hello-log-sync' is used for generating `logName`

#### Actual results:
![image](https://user-images.githubusercontent.com/32197676/148850721-b924e9cb-93e8-4fca-aeb0-82381b8bcd43.png)
1. valid `logName` is a part of the `jsonPayload` for some reason
2. `data` ended up in `message`, but I didn't specified any messages
3. `labels` looks ok
4. standard `logName` is generated instead

#### Notes:
**What if I will switch to `logging.log` instead?**
I works A-OK, but please check the next question
 
**Why I'm using logSync?**
I want to have `executionId` label autogenerated which is not available in `logging.log`
  
**can I use `console.log` instead?**
Yes, but from what I know, I can't specify `log name` and affect `logName` using `console.log` which is critical for me

anton-yarchuk:
Thanks, @losalex for the reply

A bit more context, MB it will help:
I came from [another issue](https://github.com/googleapis/nodejs-logging/issues/591), which is related to the lack of `executionId` when you are using regular `Log` in background tasks. This issue is quite old btw. 

At the very end of the discussion (the very last comment), @0xSage suggested to use `LogSync` instead or regular `Log`. 
Indeed, with `LogSync` `executionId` label is added, but if you will check the code in his comment you will notice that  he had the same issue with duplicated `logName`:

![image](https://user-images.githubusercontent.com/32197676/150518129-b3871b5e-c780-47f0-8e67-7b1d58c9e6b8.png)

losalex:
[...]

PR #1223, opened 2022-01-27
fix: logSync ignores log_id when it generates logName field

losalex:
Adding an ability to wrap all entries logged in LogSync.write with [write requests](https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write).
This fix is controlled by `LogSyncOptions.useWriteRequest` parameter which be default is false

Fixes #<[1207](https://github.com/googleapis/nodejs-logging/issues/1207)> 🦕

### DIFF
--- a/src/log-sync.ts
+++ b/src/log-sync.ts
@@ -30,6 +30,11 @@ import {
   WriteOptions,
 } from './utils/log-common';
 
+export interface LogSyncOptions {
+  // A flag indicating if written log entries should be wrapped with write request. See {@link  https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/write|entries.write}
+  useWriteRequest?: boolean;
+}
+
 /**
  * A logSync is a named collection of entries in structured log format. In Cloud
  * Logging, structured logs refer to log entries that use the jsonPayload field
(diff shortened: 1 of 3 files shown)


src/log-sync.ts
```diff
@@ -425,7 +437,13 @@ class LogSync implements LogSeverityFunctions {
...
+        if (this.options.useWriteRequest) {
+          const reqOpts = {
+            logName: this.formattedName_,
+            entries: arrify(entry) as StructuredJson[],
+          };
+          this.transport.write(JSON.stringify(reqOpts) + '\n');
+        } else this.transport.write(JSON.stringify(entry) + '\n');
       }
     } catch (err) {
       // Ignore errors (client libraries do not panic).
```
savijatv, line 449: Can we modify this comment to a more generic statement for errors and move it above line#448?

losalex: Sure thing - any suggestions how to make it more generic?

minherz:
@losalex since customizing the log name is not supported when printing structured log to stdout, i propose to document the existing `logging.logSync()` method and explain that the logname will be pushed into the `jsonPayload`.
In addition to provide an overload for this method that does not accept parameters.
We can mark the current method as deprecated to avoid ambiguity.

Issue #1185, opened 2021-11-15
Regular error showing Total timeout of API google.logging.v2.LoggingServiceV2 exceeded 600000 milliseconds


yhernesto:
Hi, Im getting a similar error:

```
C:\Users\Lenovo\myproject\node_modules\google-gax\build\src\normalCalls\retries.js:66
                const error = new googleError_1.GoogleError(`Total timeout of API ${apiName} exceeded ${retry.backoffSettings.totalTimeoutMillis} milliseconds before any response was received.`);
                              ^

GoogleError: Total timeout of API google.logging.v2.LoggingServiceV2 exceeded 60000 milliseconds before any response was received.
    at repeat (C:\Users\Lenovo\myproject\node_modules\google-gax\build\src\normalCalls\retries.js:66:31)
    at Timeout._onTimeout (C:\Users\Lenovo\myproject\node_modules\google-gax\build\src\normalCalls\retries.js:101:25)
    at listOnTimeout (node:internal/timers:557:17)
    at processTimers (node:internal/timers:500:7) {
  code: 4
}
```
Im using:
"@google-cloud/language": "^4.2.8",
"@google-cloud/logging": "^9.6.2",
"@google-cloud/storage": "^5.15.0",

with: "google-gax": "^2.28.1".

I'm running a nestjs app and using [Google Logging for NodeJS](https://cloud.google.com/logging/docs/setup/nodejs)

Thanks a lot in advance for your help!

_Originally posted by @yhernesto in https://github.com/googleapis/nodejs-logging/issues/972#issuecomment-968173656_

varungbt:
seeing this error too in our logs from time to time.

yaakovfeldman:
I get this error too if a request came to my app (in local development) when the internet was disconnected. This is obviously not likely to occur in production, but it is possibly unexpected behaviour for the app to crash whenever the logging library is unable to reach the GCP endpoint?

niekcandaele:
Also seeing this error, using `"@google-cloud/logging-winston": "^4.1.1"`. The app has internet connectivity at all times but still the error happens

jyu-hca:
Hi, I'm also getting this error in google cloud function.

node v12,
"@google-cloud/logging": "^9.2.2",
"@google-cloud/logging-winston": "^4.0.4",
pub/sub cloud function

The cloud function finished with status 'ok' at 18:40:27, and then it arises an `Unhandled rejection` and timeout error at 18:41:45

[...]

PR #1225, opened 2022-01-31
fix: Regular error showing Total timeout of API google.logging.v2.LoggingServiceV2 exceeded 600000 milliseconds

losalex:
This is a proposed fix to accept a global callback for `Log.delete` and `Log.write` methods through `LogOptions`

Fixes #[<1185>](https://github.com/googleapis/nodejs-logging/issues/1185) 🦕

### DIFF
--- a/.readme-partials.yml
+++ b/.readme-partials.yml
@@ -125,3 +125,4 @@ body: |-
 
     const log = logging.log('my-log', options);
   ``` 
+  See the full sample in `writeLogWithCallback` function [here](https://github.com/googleapis/nodejs-logging/blob/master/samples/logs.js).
--- a/.readme-partials.yml
+++ b/.readme-partials.yml
@@ -88,7 +88,8 @@ body: |-
   ```
 
   However, awaiting for every `log.write` or `log.delete` calls may introduce delays which could be avoided by 
-  simply adding a callback like in example below:
+  simply adding a callback like in example below - this way the log entry can be queued for processing and code
+  execution will continue without further delays, so callback will be called once operation is complete:
 
   ```js
       // Asynchronously write the log entry and handle respone or any errors in provided callback
--- a/.readme-partials.yml
+++ b/.readme-partials.yml
@@ -76,7 +76,7 @@ body: |-
 
   The `Log` class provide users the ability to write and delete logs asynchronously. However, there are cases when log entries
   cannot be written or deleted and error is thrown - if error is not handled properly, it could crash the application.
-  One possible way to catch all errors is to `await` the log write/delete calls and wrap it with `try/catch` like in example below:
+  One possible way to catch the error is to `await` the log write/delete calls and wrap it with `try/catch` like in example below:
 
   ```js
       // Write log entry and and catch any errors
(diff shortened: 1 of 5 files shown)


src/log.ts
```diff
@@ -60,6 +60,7 @@ export interface LogOptions {
   removeCircular?: boolean;
   maxEntrySize?: number; // see: https://cloud.google.com/logging/quotas
   jsonFieldsToTruncate?: string[];
+  defaultWriteDeleteCallback?: ApiResponseCallback;
```
summer-ji-eng, line 63: Is this global callback only for WriteDelete? I thought you want to have a global callback for all methods in Log class in manual layer.

losalex: It is for `write()` and `delete()` APIs which are the only one's used to update logs. Just wanted to distinguish it to make sure we do not confuse developers

summer-ji-eng (approved):

src/log.ts
minherz: a small correction:`delete()` does not update logs, it deletes all log entries for a provided log name as a result of the delete log operation while `write()` creates one or more log entries.

i am not sure about the "global" scope. the scope is bound by the `Log` instance, isn't it?

minherz (commented):
From the issue (#1185) it is not clear what developers except to get. The complains sound like they do not want an application to crash in a case the retries fail to ingest the log.
In this respect, the fix does not resolve the problem if the callback is not provided.

src/log.ts
losalex: > 

Based on user's feedback in a bug, the suggestion was: "I assume a lot of users will want to set up some sort of global error catching for this particular error."
Do you have specific proposal in mind how to fix this?

losalex:
> From the issue (#1185) it is not clear what developers except to get. The complains sound like they do not want an application to crash in a case the retries fail to ingest the log. In this respect, the fix does not resolve the problem if the callback is not provided.

Based on user's feedback in a bug, the suggestion was: "I assume a lot of users will want to set up some sort of global error catching for this particular error."
Do you have specific proposal in mind how to fix this?

minherz:
The quoted suggestion comes from one out of four people who describe scenarios that generally caused by failure in retrying mechanism due to transient connectivity error or because the ran their application in the not connected environment.

[One of the cases](https://github.com/googleapis/nodejs-logging/issues/1185#issuecomment-982324947) resulted in printing the error to STDOUT where it was catched by a logging agent and eventually reported to Cloud Logging backend.

In my opinion the problem is when this error crashes the application. It is a good idea to provide a way to define the error callback. I suggest to add a default error handling if nothing is provided, so the application would not crash.

losalex:
> The quoted suggestion comes from one out of four people who describe scenarios that generally caused by failure in retrying mechanism due to transient connectivity error or because the ran their application in the not connected environment.
> 
> [One of the cases](https://github.com/googleapis/nodejs-logging/issues/1185#issuecomment-982324947) resulted in printing the error to STDOUT where it was catched by a logging agent and eventually reported to Cloud Logging backend.
> 
> In my opinion the problem is when this error crashes the application. It is a good idea to provide a way to define the error callback. I suggest to add a default error handling if nothing is provided, so the application would not crash.

I believe we cannot provide default callback since client may call `log.write` with `await` and put the call inside `try/catch` - this way default callback from code will "swallow" any thrown error, thus catch becomes to be redundant. Worth mentioning that it is eventually will violate our contract with users who does `await` with `try/catch`.

.readme-partials.yml
```diff
@@ -71,3 +71,58 @@ body: |-
...
+      try {
+        await log.write(entry);
+      } catch (err) {
+        console.log('Error is: ' + err);
+      }
+  ```
+
+  However, awaiting for every `log.write` or `log.delete` calls may introduce delays which could be avoided by 
+  simply adding a callback like in example below - this way the log entry can be queued for processing and code
+  execution will continue without further delays, so callback will be called once operation is complete:
```
keenan-devrel, line 92: in the example below. This way the log entry can be queued for processing and code
  execution will continue without further delays.  The callback will be called once the operation is complete:

losalex: Thanks!

PR #1261, opened 2022-04-29
feat: Add support for library instrumentation

losalex:
This feature provides an ability to log extra entry with diagnostics structure which contains logging library information. Such entry is logged only once when first entry is written by a process using logging library.

Fixes #[1286](https://github.com/googleapis/nodejs-logging/issues/1286) 🦕

### DIFF
--- a/src/log.ts
+++ b/src/log.ts
@@ -14,13 +14,13 @@
  * limitations under the License.
  */
 
-import arrify = require('arrify');
 import {callbackifyAll} from '@google-cloud/promisify';
 import * as dotProp from 'dot-prop';
 import * as extend from 'extend';
 import {CallOptions} from 'google-gax';
 import {GetEntriesCallback, GetEntriesResponse, Logging} from '.';
 import {Entry, EntryJson, LogEntry} from './entry';
+import {populatedInstrumentationInfo} from './utils/instrumentation';
 import {
   LogSeverityFunctions,
   assignSeverityToEntries,
@@ -964,8 +964,8 @@ class Log implements LogSeverityFunctions {
     await this.logging.setProjectId();
     this.formattedName_ = formatLogName(this.logging.projectId, this.name);
     const resource = await this.getOrSetResource(options);
-    // Extract & format additional context from individual entries.
-    const decoratedEntries = this.decorateEntries(arrify(entry) as Entry[]);
+    // Extract & format additional context from individual entries. Make sure to add instrumentation info
+    const decoratedEntries = this.decorateEntries(populatedInstrumentationInfo(entry));
     this.truncateEntries(decoratedEntries);
     // Clobber `labels` and `resource` fields with WriteOptions from the user.
     const reqOpts = extend(
--- a/src/log.ts
+++ b/src/log.ts
@@ -20,7 +20,10 @@ import * as extend from 'extend';
 import {CallOptions} from 'google-gax';
 import {GetEntriesCallback, GetEntriesResponse, Logging} from '.';
 import {Entry, EntryJson, LogEntry} from './entry';
-import {populatedInstrumentationInfo} from './utils/instrumentation';
+import {
+  populatedInstrumentationInfo,
+  getInstrumentationInfoStatus,
+} from './utils/instrumentation';
 import {
   LogSeverityFunctions,
   assignSeverityToEntries,
@@ -959,7 +962,15 @@ class Log implements LogSeverityFunctions {
     entry: Entry | Entry[],
     opts?: WriteOptions | ApiResponseCallback
   ): Promise<ApiResponse> {
+    const isInfoAdded = getInstrumentationInfoStatus();
     const options = opts ? (opts as WriteOptions) : {};
+    // If instrumentation info was not added, means that this is first time
+    // log entry is written and that the instrumentation log entry could be
+    // generated for this request. If yes, then make sure we set partialSuccess, so entire
+    // request will make it through and only oversized entries will be dropped
+    if (!isInfoAdded) {
+      options.partialSuccess = true;
+    }
     // Extract projectId & resource from Logging - inject & memoize if not.
     await this.logging.setProjectId();
     this.formattedName_ = formatLogName(this.logging.projectId, this.name);
--- a/src/log.ts
+++ b/src/log.ts
@@ -21,7 +21,7 @@ import {CallOptions} from 'google-gax';
 import {GetEntriesCallback, GetEntriesResponse, Logging} from '.';
 import {Entry, EntryJson, LogEntry} from './entry';
 import {
-  populatedInstrumentationInfo,
+  populateInstrumentationInfo,
   getInstrumentationInfoStatus,
 } from './utils/instrumentation';
 import {
@@ -977,7 +977,7 @@ class Log implements LogSeverityFunctions {
     const resource = await this.getOrSetResource(options);
     // Extract & format additional context from individual entries. Make sure to add instrumentation info
     const decoratedEntries = this.decorateEntries(
-      populatedInstrumentationInfo(entry)
+      populateInstrumentationInfo(entry)
     );
     this.truncateEntries(decoratedEntries);
     // Clobber `labels` and `resource` fields with WriteOptions from the user.
--- a/src/log.ts
+++ b/src/log.ts
@@ -965,7 +965,9 @@ class Log implements LogSeverityFunctions {
     this.formattedName_ = formatLogName(this.logging.projectId, this.name);
     const resource = await this.getOrSetResource(options);
     // Extract & format additional context from individual entries. Make sure to add instrumentation info
-    const decoratedEntries = this.decorateEntries(populatedInstrumentationInfo(entry));
+    const decoratedEntries = this.decorateEntries(
+      populatedInstrumentationInfo(entry)
+    );
     this.truncateEntries(decoratedEntries);
     // Clobber `labels` and `resource` fields with WriteOptions from the user.
     const reqOpts = extend(
--- a/src/utils/instrumentation.ts
+++ b/src/utils/instrumentation.ts
@@ -15,23 +15,27 @@
  */
 
 import arrify = require('arrify');
+import path = require('path');
 import {google} from '../../protos/protos';
 import {Entry} from '../entry';
-import version = require('../../package.json');
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+declare const global: {[index: string]: any};
 
 // The global variable keeping track if instrumentation record was already written or not.
 // The instrumentation record should be generated only once per process and contain logging
 // libraries related info.
-// eslint-disable-next-line @typescript-eslint/no-explicit-any
-declare const global: {[index: string]: any};
 global.instrumentationAdded = false;
 
+// The variable to hold cached library version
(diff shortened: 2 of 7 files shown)


src/utils/instrumentation.ts
```diff
@@ -0,0 +1,201 @@
...
+export function getInstrumentationInfoStatus() {
+  return global.instrumentationAdded;
+}
+
+/**
+ * This method helps to populate entries with instrumentation data
+ * @param entry {Entry} The entry or array of entries to be populated with instrumentation info
+ * @returns {Entry} Array of entries which contains an entry with current library instrumentation info
+ */
+export function populatedInstrumentationInfo(entry: Entry | Entry[]): Entry[] {
```
arbrown, line 54: Is this function supposed to be 'populateInstrumentationInfo' (without the d at the end of populate?) If not, why is it 'populated' (past tense?)

src/utils/instrumentation.ts
```diff
@@ -0,0 +1,201 @@
...
+  // Update the flag indicating that instrumentation entry was already added once,
+  // so any subsequent calls to this method will not add a separate instrumentation log entry
+  let isWritten = global.instrumentationAdded;
+  global.instrumentationAdded = true;
+  const entries: Entry[] = [];
+  if (entry) {
```
arbrown, line 60: Could you skip this block if `isWritten` is already true? Or is it necessary to validate and update subsequent logs with additional instrumentation info?

src/utils/instrumentation.ts
```diff
@@ -0,0 +1,201 @@
...
+ * The helper function which checks if given InstrumentationInfo is valid.
+ * @param info {InstrumentationInfo} The info to be validated.
+ * @returns true if given info is valid, false otherwise
+ */
+function isValidInfo(info: InstrumentationInfo) {
+  if (
+    !info ||
+    !info.name ||
+    !info.version ||
+    !info.name.startsWith(NODEJS_LIBRARY_NAME_PREFIX)
```
arbrown, line 189: Why do you validate that every entry has the same `name` field?  What is the use case for keeping 3 identical entries? Are there any cases where this could come pre-populated with another library name and we would want to preserve that information (eg a nodejs logging agent that processes logs written in a Go microservice)?

src/utils/instrumentation.ts
```diff
@@ -0,0 +1,201 @@
...
+function truncateValue(value: string) {
+  if (value && value.length > maxDiagnosticValueLen) {
+    return value.substring(0, maxDiagnosticValueLen).concat('*');
+  }
+  return value;
+}
+
+/**
+ * The helper function to retrieve current library version from 'package.json' file. Note that
+ * since we use {path.resolve}, the search for 'package.json' could be impacted by current working directory.
```
arbrown, line 164: What is the failure case for path.resolve and different current working directories?  Does path.resolve use the working directory of the process or the source code file as it appears here?

src/log.ts
```diff
@@ -959,13 +962,23 @@ class Log implements LogSeverityFunctions {
...
   ): Promise<ApiResponse> {
+    const isInfoAdded = getInstrumentationInfoStatus();
     const options = opts ? (opts as WriteOptions) : {};
+    // If instrumentation info was not added, means that this is first time
+    // log entry is written and that the instrumentation log entry could be
+    // generated for this request. If yes, then make sure we set partialSuccess, so entire
+    // request will make it through and only oversized entries will be dropped
+    if (!isInfoAdded) {
+      options.partialSuccess = true;
+    }
```
minherz, line 973: I understand the reasoning but a question remains how does the library know that the instrumentation log entry was successfully ingested?

src/utils/instrumentation.ts
```diff
@@ -0,0 +1,201 @@
...
+
+// The variable to hold cached library version
+let libraryVersion: string;
+
+// Max length for instrumentation library name and version values
+const maxDiagnosticValueLen = 14;
+
+export const DIAGNOSTIC_INFO_KEY = 'logging.googleapis.com/diagnostic';
+export const INSTRUMENTATION_SOURCE_KEY = 'instrumentation_source';
+export const NODEJS_LIBRARY_NAME_PREFIX = 'nodejs';
```
minherz, line 38: nit: is it approved language code? can we add a link to the list of code references (e.g. in gax/grpc)

src/utils/instrumentation.ts
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * This method helps to populate entries with instrumentation data
+ * @param entry {Entry} The entry or array of entries to be populated with instrumentation info
+ * @returns {Entry} Array of entries which contains an entry with current library instrumentation info
+ */
+export function populatedInstrumentationInfo(entry: Entry | Entry[]): Entry[] {
+  // Update the flag indicating that instrumentation entry was already added once,
+  // so any subsequent calls to this method will not add a separate instrumentation log entry
+  let isWritten = global.instrumentationAdded;
+  global.instrumentationAdded = true;
```
minherz, line 58: nit: if the ingestion fails does the flag remains set?

src/utils/instrumentation.ts
```diff
@@ -0,0 +1,201 @@
...
+  }
+  return true;
+}
+
+/**
+ * The helper method used for tests to reset a status if instrumentation already written.
+ */
+export function testResetInstrumentationStatus() {
+  global.instrumentationAdded = false;
+}
```
minherz, line 201: nit: i hope there is a method to avoid placing test related code into the body of the exported package. maybe worth to consult with Benjamin

minherz (commented):
good work. maybe it worth to bring someone from Yoshi team for additional review.

src/utils/instrumentation.ts
losalex: Ah, good catch - thank you! Fixing

src/utils/instrumentation.ts
losalex: Yeah, we should validate if subsequent logs are written for example by other library. Also, higher level library could decide to write such a log entry more frequently, so in this case we could just inspect the load and add relevant info.

src/utils/instrumentation.ts
losalex: I validate the prefix only - every library using this feature should start with 'nodejs' prefix

src/utils/instrumentation.ts
losalex: Yes, it uses CWD as a base path. I had to make this change due to system tests defined in [install.ts](https://github.com/googleapis/nodejs-logging/blob/main/system-test/install.ts) - apparently those tests build a package and check if such package can be installed. Those tests were initially failing for me simply because I used relative path to retrieve a version, e.g. by running `require('../../package.json').version`.

arbrown (approved):

src/log.ts
losalex: Library doesn't know - we do send the log entry and rely on retries to deliver an entry. The entry containing instrumentation info supposed to be not oversized (e.g. library creates this entry and keep track on data added to the entry). partialSuccess is added to ensure that entry is not dropped if request contains oversized entries. I will take a look to see if we can also track the request success/failure and perhaps will add a logic to reset global flag to resend instrumentation data.

src/utils/instrumentation.ts
losalex: Yes, it remains - we rely on retries to deliver an entry. Will look to see if I can track the request with callback, however it might be dangerous - what if error still returned for request containing oversized entries despite a fact that some entries were actually committed?

src/utils/instrumentation.ts
losalex: It is - I used a reference from [OTel](https://github.com/open-telemetry/opentelemetry-specification/tree/a6cfbc92d36aa019c95fc2d9f023f5de5d570937/specification/resource/semantic_conventions#telemetry-sdk) telemetry for this

src/utils/instrumentation.ts
losalex: I asked of course, but seems Nodejs do not have magic solutions here. I am thinking to make this method to be non-test related, WDYT?

src/utils/instrumentation.ts
losalex: I checked and indeed, the error for entire request will be returned in case when some of entries were dropped... We probable should address this in separate PR - it requires deeper understanding and for now I believe retries should do the job.

test/instrumentation.ts
```diff
@@ -0,0 +1,184 @@
...
+const VERSION_TEST = '1.0.0';
+const LONG_VERSION_TEST = VERSION_TEST + '.0.0.0.0.0.0.0.0.11.1.1-ALPHA';
+
+describe('instrumentation_info', () => {
+  beforeEach(() => {
+    instrumentation.resetInstrumentationStatus();
+  });
+
+  it('should generate library info properly by default', () => {
+    console.log('The version is: ' + instrumentation.getNodejsLibraryVersion());
```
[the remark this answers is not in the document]
summer-ji-eng: Or you could use `assert` to ensure the node.js version you want.

test/instrumentation.ts
```diff
@@ -0,0 +1,184 @@
...
+      instrumentation.NODEJS_LIBRARY_NAME_PREFIX
+    );
+  });
+
+  it('should add instrumentation info to existing list', () => {
+    const dummyEntry = createEntry(NODEJS_TEST, VERSION_TEST);
+    const entries = instrumentation.populateInstrumentationInfo(dummyEntry);
+    assert.equal(entries.length, 1);
+    assert.equal(
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
summer-ji-eng, line 75: To get rid of this linter, maybe this help:
```
entries.forEach(entry => {
 assert(entry instanceof Entry);
 // do the rest of equal check;
```

summer-ji-eng (approved):
But familiar with the business, but overall LGTM and left couple comments. Thank you:)

test/instrumentation.ts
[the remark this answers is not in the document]
losalex: Oh, good catch - there is already an assertion below, this is debugging leftover.

src/utils/instrumentation.ts
```diff
@@ -0,0 +1,205 @@
...
+  // Validate the libraryName first and make sure it starts with 'nodejs' prefix.
+  if (!libraryName || !libraryName.startsWith(NODEJS_LIBRARY_NAME_PREFIX)) {
+    libraryName = NODEJS_LIBRARY_NAME_PREFIX;
+  }
+  const entry = new Entry(
+    {
+      severity: google.logging.type.LogSeverity.INFO,
+    },
+    {
+      [DIAGNOSTIC_INFO_KEY]: {
```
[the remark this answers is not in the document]
losalex: This is a format used when you have some namespace prefix like for example `logging.googleapis.com/diagnostic`

test/instrumentation.ts
losalex: Seems that I had that conversion as leftover from previously written code - removed casting completely, thanks!

rr923:
Is it possible to disable "library instrumentation"?

rr923:
Or how can I hide this kind of entries in the log explorer? There are just too many of them.

Issue #1301, opened 2022-07-02
Logging to stdout in Cloud Run creates a JSON object as "message"


losalex:
When enabling structured logging to STDOUT, received structured JSON payload containing an extra "message" key with the original logged message, as well as a "metadata" key of an empty object.
See related bug [704](https://github.com/googleapis/nodejs-logging-winston/issues/704)

Allezxandre:
I have attempted a fix on this branch: https://github.com/prgm-dev/nodejs-logging/commit/a220caea5a4f6e77038b91801889dc308175b539

Been using it in production for about 2 weeks now, but not sure it's the right fix.

Will try to open a PR with these changes this week, but as @losalex remarked in https://github.com/googleapis/nodejs-logging-winston/issues/704, this might be a breaking change for users relying on this outside of Cloud Run.

losalex:
> I have attempted a fix on this branch: [prgm-dev@a220cae](https://github.com/prgm-dev/nodejs-logging/commit/a220caea5a4f6e77038b91801889dc308175b539)
> 
> Been using it in production for about 2 weeks now, but not sure it's the right fix.
> 
> Will try to open a PR with these changes this week, but as @losalex remarked in [googleapis/nodejs-logging-winston#704](https://github.com/googleapis/nodejs-logging-winston/issues/704), this might be a breaking change for users relying on this outside of Cloud Run.

Thank you @Allezxandre - this is very helpful! Indeed - I am looking into your change now and it is very useful, so I am planning also to add the parameter to control this change with default to be set to "off" so it wont break existing clients. Later when we have a major release we might consider turning it "on" by default - let me know if you have any comments

losalex:
Will work to propagate this fix to winston/bunyan as well

PR #1305, opened 2022-07-08
fix: Logging to stdout in Cloud Run creates a JSON object as "message"

losalex:
This is a fix to provide a configurable way to disable "message" wrapper added for structured log data when LogSync is used for logged entries

Fixes #<[1301](https://github.com/googleapis/nodejs-logging/issues/1301)> 🦕

### DIFF
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -275,12 +275,19 @@ class Entry {
         ? meta.traceSampled
         : undefined;
     // Format log payload.
-    let data =
-      meta.textPayload || meta.jsonPayload || meta.protoPayload || undefined;
+    const data =
+      this.data ||
+      meta.textPayload ||
+      meta.jsonPayload ||
+      meta.protoPayload ||
+      undefined;
     if (useMessageField) {
-      entry.message = this.data || data;
+      /** If useMessageField is set, we add the payload to {@link StructuredJson#message} field.*/
+      entry.message = data;
     } else {
-      data = this.data || data;
+      /** useMessageField is false, we add the structured payload to {@link StructuredJson} key-value map.
+       * It could be especially useful for serverless environments like Cloud Run/Functions when stdout transport is used.
+       * Note that text still added to {@link StructuredJson#message} field for text payload since it does not have fields within. */
       if (data !== undefined && data !== null) {
         if (this.isObject(data)) {
           entry = extend(true, {}, entry, data);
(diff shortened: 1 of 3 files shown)


src/entry.ts
```diff
@@ -243,7 +245,7 @@ class Entry {
    * Serialize an entry to a standard format for any transports, e.g. agents.
    * Read more: https://cloud.google.com/logging/docs/structured-logging
    */
-  toStructuredJSON(projectId = '') {
+  toStructuredJSON(projectId = '', useMessageField = true) {
```
daniel-sanche, line 248: I'm a bit confused by this. Will end users be setting `useMessageField` manually? I don't think other languages work this way

src/entry.ts
```diff
@@ -273,9 +275,22 @@ class Entry {
...
         : undefined;
     // Format log payload.
-    entry.message =
+    let data =
       meta.textPayload || meta.jsonPayload || meta.protoPayload || undefined;
-    entry.message = this.data || entry.message;
+    if (useMessageField) {
+      entry.message = this.data || data;
+    } else {
+      data = this.data || data;
```
daniel-sanche, line 283: nit: this block could probably use some comments

src/entry.ts
losalex: `useMessageField` will be set through options for those users who does not want to have extra "message" field in `jsonPayload` while running in Cloud Run/Functions and using `stdout` logging. Indeed, other languages dont have this bug - I am fixing this just to make sure not to break existing users which use `LogSync` and already get used to the extra "message" field in `jsonPayload`.

losalex: Yep - for next major release we will turn this flag to be false by default

src/entry.ts
losalex: Added

src/entry.ts
```diff
@@ -273,9 +275,22 @@ class Entry {
...
+      if (data !== undefined && data !== null) {
+        if (this.isObject(data)) {
+          entry = extend(true, {}, entry, data);
+        } else if (typeof data === 'string') {
+          entry.message = data;
+        } else {
+          entry.message = JSON.stringify(data);
```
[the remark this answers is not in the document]
losalex: Yep, mostly - I will definitely add non-text simple test (e.g. numeric)

daniel-sanche (approved):

PR #1310, opened 2022-08-18
fix: Instrumentation version contains package.json content

losalex:
In few cases when we retrieve the package version by calling [getNodejsLibraryVersion](https://github.com/googleapis/nodejs-logging/blob/42859ab634721cb57ca76960f8a8aa5f6dd3a987/src/utils/instrumentation.ts#L168) the content from package.json is returned. This fix tries to retrieve version from JSON and returns 'unknown' value if version cannot be retrieved


src/utils/instrumentation.ts
```diff
@@ -154,6 +154,22 @@ function validateAndUpdateInstrumentation(
  * @returns {string} The truncated value.
  */
 function truncateValue(value: string, maxLen: number) {
```
ruyadorno, line 156: Mind it has been a while since I've touched a TypeScript codebase but I'm curious about this `value` parameter here. Should it maybe be tweaked to indicate that it also accepts an object with a value property? Something similar to: https://github.com/googleapis/nodejs-logging/blob/42859ab634721cb57ca76960f8a8aa5f6dd3a987/src/index.ts#L440

test/instrumentation.ts
```diff
@@ -51,6 +51,27 @@ describe('instrumentation_info', () => {
...
+      instrumentation.NODEJS_LIBRARY_NAME_PREFIX
+    );
+    assert.equal(
+      entry.data?.[instrumentation.DIAGNOSTIC_INFO_KEY]?.[
+        instrumentation.INSTRUMENTATION_SOURCE_KEY
+      ]?.[0]?.[VERSION],
+      'unknown'
+    );
+  });
+
```
ruyadorno, line 74: It would be nice to also have a test for retrieving the `version` value from a `package.json` file here.

ruyadorno (commented):
Just an optional suggestion to add a test, otherwise implementation LGTM 👍

src/utils/instrumentation.ts
losalex: Makes sense, changed - thanks @ruyadorno!

test/instrumentation.ts
losalex: Yep, there is one - see [here](https://github.com/googleapis/nodejs-logging/blob/42859ab634721cb57ca76960f8a8aa5f6dd3a987/test/instrumentation.ts#L35)

ruyadorno: oh that's great!

ruyadorno (approved):

PR #1311, opened 2022-08-18
fix: maxResults in getEntries looks broken

### LINKED ISSUE #1277: maxResults in getEntries looks broken

1) Is this a client library issue or a product issue?
This is the client library for . We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [ Support page]() to reach the most relevant engineers.
`Client Library issue`
2) Did someone already solve this?
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
`Nope`
3) Do you have a support contract?
Please create an issue in the [support console](https://cloud.google.com/support/) to ensure a timely response.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: linux
  - Node.js version: 16.13.2
  - npm version: 8.1.2
  - `@google-cloud/logging` version: ^9.9.0"

#### Steps to reproduce

  1. 
  ```js
  const loggingClient = new Logging();
  const [entries] = await loggingClient.getEntries({
        autoPaginate: false,
        //filter: filters.join('\n'),
        maxResults: 5,
    })
    
   expect(entries.length).toBe(5)
  ```
  2.  output:
  ```
      Expected: 5
    Received: 50
  ```

Making sure to follow these steps will guarantee the quickest resolution possible.

Thanks!

losalex:
The maxResults parameter was not propagated properly

Fixes [#1277](https://github.com/googleapis/nodejs-logging/issues/1277) 🦕


src/index.ts
```diff
@@ -592,12 +592,20 @@ class Logging {
...
       options!.gaxOptions
     );
+    if (options?.maxResults) {
+      gaxOptions = extend(
+        {
+          maxResults: options.maxResults,
+        },
+        gaxOptions
+      );
+    }
```
minherz, line 608: nit: why not to update only the `maxResults` field? Unless it is a case of the dynamic structure when the code operates with typeless architecture?

minherz (approved):
lgtm. one nit question/suggestion for your consideration.

src/index.ts
losalex: Indeed - we operate here on typeless structure

Issue #1266, opened 2022-05-07
Doesn't work as described in README: `log.logSync(logName).write(log.entry(metadata, payload))`


chriscalo:
The behavior of `log.logSync(logName).write(log.entry(metadata, payload))` doesn't match what's described in the README.

#### Environment details

  - OS: Google App Engine Standard Environment for Node.js
  - Node.js version: 16
  - npm version: 
  - `@google-cloud/logging` version: 9.8.3

I've created a reproduction at https://github.com/chriscalo/logsync-bug and copied the full details here:

## Steps:
1. Deploy to App Engine Standard environment:

``` sh
gcloud app deploy app.yaml --project=<PROJECT_ID>
```

2. Go to [console.cloud.google.com/logs/](http://console.cloud.google.com/logs/) to view output

## Expected:

Here are the first two calls to `log.write()` in `index.js`:

``` js
const metadata = {
  resource: {
    type: "gae_app",
    labels: {
      module_id: process.env.GAE_SERVICE,
      version_id: process.env.GAE_VERSION,
    },
  },
};

log.write(log.entry(metadata, "A text payload"));
log.write(log.entry(metadata, { message: "A JSON payload" }));
```

The expected output is something like: (irrelevant details removed)

``` json
[
  {
    "textPayload": "A text payload",
    "resource": {
      "type": "gae_app",
      "labels": {
        "version_id": "logsync-bug",
        "module_id": "default",
      }
    },
    "timestamp": "2022-05-06T23:13:23.002Z",
    "logName": "projects/REDACTED_PROJECT_ID/logs/my_log"
  },
  {
    "jsonPayload": {
      "message": "A JSON payload"
    },
    "resource": {
      "type": "gae_app",
      "labels": {
        "module_id": "default",
        "version_id": "logsync-bug"
      }
    },
    "timestamp": "2022-05-06T23:13:23.004Z",
    "logName": "projects/REDACTED_PROJECT_ID/logs/my_log",
    "receiveTimestamp": "2022-05-06T23:13:23.270510336Z"
  }
]
```

## Actual:

The actual output looks like the following: (irrelevant details removed)

``` json
[
  {
    "jsonPayload": {
      "message": "A text payload",
      "timestamp": "2022-05-06T23:13:23.002Z",
      "logName": "projects/REDACTED_PROJECT_ID/logs/my_log",
      "resource": {
        "type": "gae_app",
        "labels": {
          "module_id": "default",
[...]

PR #1328, opened 2022-09-02
fix: Correction for timestamp and instrumentation record severity fields format

losalex:
It seems that `timestamp` format for log records printed to stdout are in wrong format - the right format is defined here: [timestamp-processing](https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing). Worth mentioning that changing this format should be safe despite a fact that `StructuredJson` defined as external - it is used internally and I believe it is safe to change format of `timestamp`.
Also instrumentation record was setting `severity` as number and since it should be with `DEFAULT` severity, removing it from instrumentation record.

Related to # [1266](https://github.com/googleapis/nodejs-logging/issues/1266) 🦕

### DIFF
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -22,7 +22,7 @@ import {
   objToStruct,
   structToObj,
   zuluToDateObj,
-  toNanoSecondsObj,
+  toNanosAndSecondsObj,
 } from './utils/common';
 import {
   makeHttpRequestData,
@@ -81,7 +81,7 @@ export interface StructuredJson {
   message?: string | object;
   httpRequest?: object;
   // Based on https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing, the
-  // timestamp should be in nanoseconds format.
+  // timestamp should be in nanos and seconds format.
   timestamp?: Timestamp;
   [INSERT_ID_KEY]?: string;
   [OPERATION_KEY]?: object;
@@ -221,7 +221,7 @@ class Entry {
     }
     // Format log timestamp
     if (entry.timestamp instanceof Date) {
-      entry.timestamp = toNanoSecondsObj(entry.timestamp);
+      entry.timestamp = toNanosAndSecondsObj(entry.timestamp);
     } else if (typeof entry.timestamp === 'string') {
       entry.timestamp = zuluToDateObj(entry.timestamp);
     }
@@ -302,7 +302,7 @@ class Entry {
     }
     // Format timestamp
     if (meta.timestamp instanceof Date) {
-      entry.timestamp = toNanoSecondsObj(meta.timestamp);
+      entry.timestamp = toNanosAndSecondsObj(meta.timestamp);
     }
     // Format httprequest
     const req = meta.httpRequest;
--- a/src/entry.ts
+++ b/src/entry.ts
@@ -18,7 +18,12 @@
 const EventId = require('eventid');
 import * as extend from 'extend';
 import {google} from '../protos/protos';
-import {objToStruct, structToObj, zuluToDateObj} from './utils/common';
+import {
+  objToStruct,
+  structToObj,
+  zuluToDateObj,
+  toNanoSecondsObj,
+} from './utils/common';
 import {
   makeHttpRequestData,
   CloudLoggingHttpRequest,
@@ -75,7 +80,9 @@ export interface StructuredJson {
   // Universally supported properties
   message?: string | object;
   httpRequest?: object;
-  timestamp?: string;
+  // Based on https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing, the
+  // timestamp should be in nanoseconds format.
+  timestamp?: Timestamp;
   [INSERT_ID_KEY]?: string;
   [OPERATION_KEY]?: object;
   [SOURCE_LOCATION_KEY]?: object;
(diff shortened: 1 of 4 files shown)


test/entry.ts
```diff
@@ -319,7 +327,11 @@ describe('Entry', () => {
       entry.metadata.traceSampled = false;
       entry.data = 'this is a log';
       const json = entry.toStructuredJSON();
-      assert(withinExpectedTimeBoundaries(new Date(json.timestamp!)));
+      assert(
+        withinExpectedTimeBoundaries(
+          new Date(nanosecondsToDate(json.timestamp!))
```
daniel-sanche, line 332: nit: I believe you can remove the `new Date`, since `nanosecondsToDate` already returns a date

test/entry.ts
```diff
@@ -56,6 +56,14 @@ function withinExpectedTimeBoundaries(result?: Date): boolean {
   return false;
 }
 
+function nanosecondsToDate(timestamp: entryTypes.Timestamp) {
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const seconds = (timestamp as any).seconds;
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const nanos = (timestamp as any).seconds;
```
daniel-sanche, line 63: is this supposed to be `.nanos` instead of `.seconds`?

test/entry.ts
```diff
@@ -345,7 +357,11 @@ describe('Entry', () => {
     it('should convert a string timestamp', () => {
       entry.metadata.timestamp = new Date();
       const json = entry.toStructuredJSON();
-      assert(withinExpectedTimeBoundaries(new Date(json.timestamp!)));
+      assert(
+        withinExpectedTimeBoundaries(
+          new Date(nanosecondsToDate(json.timestamp!))
+        )
+      );
```
daniel-sanche, line 364: It might be good to add another test to ensure that `json.timestamp` is now in the expected dictionary format.  `nanosecondsToDate()` would probably throw some errors if it wasn't, but it still might be good to be safe

src/entry.ts
```diff
@@ -75,7 +80,9 @@ export interface StructuredJson {
   // Universally supported properties
   message?: string | object;
   httpRequest?: object;
-  timestamp?: string;
+  // Based on https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing, the
```
daniel-sanche, line 83: Is there a reason you chose to use the seconds/nanos dictionary option instead of RFC 3339? It seems like it should be pretty simple to convert from a Date object to that format: https://codetogo.io/how-to-create-date-in-rfc3339-format-in-javascript/

daniel-sanche (commented):
Left a couple comments/sugg

src/entry.ts
```diff
@@ -75,7 +80,9 @@ export interface StructuredJson {
...
+  // timestamp should be in nanoseconds format.
```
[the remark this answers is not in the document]
losalex: OK - fixing comment. Please note that `Timestamp` is a dictionary containing `nanos` and `seconds` fields

src/utils/common.ts
```diff
@@ -243,3 +243,17 @@ export function zuluToDateObj(zuluTime: string) {
     nanos: nanoSecs ? Number(nanoSecs.padEnd(9, '0')) : 0,
   };
 }
+
+/**
+ * Converts Date to nanoseconds format suported by Logging.
+ * See https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing for more details
+ * @param date The date to be converted to Logging nanoseconds format
+ */
+export function toNanoSecondsObj(date: Date) {
```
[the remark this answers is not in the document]
losalex: Renaming to toNanosAndSecondsObj

test/entry.ts
losalex: Oh, right - thanks!

test/entry.ts
losalex: Ah, right! Good catch!

src/entry.ts
losalex: Please see the reference to spec I shared in description - [timestamp-processing](https://cloud.google.com/logging/docs/agent/logging/configuration#timestamp-processing).
Basically, RFC you mentioned could be placed only to `time` field - if we want to continue to use `timestamp` field, it must be nanos/seconds format. Let me know if you think we should use the `time` field with in format defined in RFC.

test/entry.ts
losalex: Thats the test which exactly testing the format - see `nanosecondsToDate` function.

src/entry.ts
daniel-sanche: I see. Note that on the backend the `time` field will become a `timestamp`, so it should be the same to the user:

> The Logging agent can receive and process time-related fields in several JSON formats.
>  If one of the following JSON timestamp representations is present in a structured record, 
> **the Logging agent strips the fields from jsonPayload and collapses them into a single 
> representation in the timestamp field in the LogEntry object**

Personally I think using the RFC format with a time field is easier, but I'm fine either way

losalex: I wanted to keep `StructuredJson.timestamp` just to make sure I am making minimal change existing structure. Also worth mentioning that `EntryJson.timestamp` already written in nanos/seconds format and seems it works today (while `StructuredJson.timestamp` is not processed today due to wrong format).

test/entry.ts
daniel-sanche: `nanosecondsToDate` isn't actually an explicit test itself though. Nothing is explicitly testing the `toNanosAndSecondsObj` function. We could be testing to make sure that `.nanos` and `.seconds` are floats instead of ints or strings, for example

daniel-sanche (approved):
I'd suggest considering using RFC3339 instead of this format, and possibly expanding the tests, but overall LGTM as is

Opened [1329](https://github.com/googleapis/nodejs-logging/issues/1329) for this suggestion. I believe the only benefit of this is readability for raw JSON payload printed to `stdout` and/or HTTP/S, hence making this P3.

test/entry.ts
losalex: I must admit it seems like overkill... I believe format tests for structures could be part of proto testing, but I think failure in `nanosecondsToDate` is good enough as indicator for wrong format.

PR #1333, opened 2022-09-08
fix: google-gax resolution

### LINKED ISSUE #1332: Incorrect dependency resolution: google-gax

1) Is this a client library issue or a product issue?
This is the client library for . We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit the [ Support page]() to reach the most relevant engineers.

2) Did someone already solve this?
  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js

3) Do you have a support contract?
Please create an issue in the [support console](https://cloud.google.com/support/) to ensure a timely response.

If the support paths suggested above still do not result in a resolution, please provide the following details.

#### Environment details

  - OS: Mac OS 12.4
  - Node.js version: 16.16
  - npm version: 8.11.0
  - `@google-cloud/logging` version: 10.8.0

#### Steps to reproduce

  1. In google-gax lib, protobuf is exported since 3.3.0 => https://github.com/googleapis/gax-nodejs/blob/main/CHANGELOG.md#330-2022-08-26
  2. A PR is created on this repo to use google-gax 3.3.0 => https://github.com/googleapis/nodejs-logging/pull/1321
  3. This is released in [v10.5.0](https://github.com/googleapis/nodejs-logging/blob/main/CHANGELOG.md#1015-2022-08-27)

The problem is that the resolution for google-gax in the package.json is 3.2.2, so people (like me), who bumped the lib from 10.1.x to 10.1.5 and later, still have the google-gax dependency installed with the version 3.2.2.

And when the lib is exporting this `module.exports = factory(require("google-gax/build/src/protobuf").protobufMinimal);`, it throws an error because it does not exist as it is only available in google-gax 3.3.0 and after.

I'll open a PR ASAP so it can be resolved :)

CorentinAndre:
Hi 👋 

This fixes #1332 🦕


package.json
```diff
@@ -97,4 +97,4 @@
   "engines": {
     "node": ">=12.0.0"
   }
-}
+}
```
losalex, line 100: Please add a missing newline to the end of file

losalex (approved):

PR #1362, opened 2022-10-28
fix: Correct an order of instrumentation entries

losalex:
Make sure that instrumentation entries are added in proper order - parent libraries first.


src/utils/instrumentation.ts
```diff
@@ -131,23 +131,23 @@ function validateAndUpdateInstrumentation(
...
   // Limit amount of entries to be up to 3
-  let count = 1;
+  let count = 0;
   for (const info of infoList) {
     if (isValidInfo(info)) {
       finalInfo.push({
         name: truncateValue(info.name, maxDiagnosticValueLen),
         version: truncateValue(info.version, maxDiagnosticValueLen),
       });
+      if (++count === 2) break;
```
daniel-sanche, line 143: nit: can this be made into a constant somewhere?

daniel-sanche (approved):

src/utils/instrumentation.ts
losalex: Sure - done

PR #1482, opened 2024-01-18
fix: retreive log entries from logging bucket

### LINKED ISSUE #1411: Not able to retreive log entries
So I have the following function here:

```
const functions = require('@google-cloud/functions-framework');
const { Storage } = require('@google-cloud/storage');
const readline = require('readline');
const { Logging } = require('@google-cloud/logging');
const storage = new Storage();

async function searchLogs(date, timeStart, timeEnd, user) {
  const logFilter = `logName="timestamp>="2023-03-01T11:55:00Z" AND timestamp<="2023-03-03T11:55:51Z"`;
  const resourceNames = ['projects/production-logs'];
  const [entries] = await logging.getEntries({ resourceNames, filter: logFilter });

  console.log(`Searching logs with filter: ${logFilter}`);
  console.log(`Found ${entries.length} entries.`);

  for (const entry of entries) {
    const message = entry.data.jsonPayload.message;
    const ipAddress = message.match(/Ip Address:\s+(\d{1,3}\.\d{1,3}\.\d{1,3}\.\d{1,3})/);
    if (ipAddress) {
      return ipAddress[1];
    }
  }

  console.log(`----`);
  return null;
}

//... rest of code
```

What I'm having an issue is I have my production logs stored in a logging bucket called "prod", however the above code only works on the project scope rather than on a per-bucket scope, so it basically can't find the logs, because it can't see the logging bucket.

Is there a way to direct it so that it sends the request to a specific logging bucket? I can't find much online, and I have found a function in the source code for the loggin API about retreiving a list of logging buckets in the project, which works, but how do I set it to use that bucket now?

Btw, I hardcoded the timestamp for now to keep it simple.

Colt45s:
I also attempted to retrieve log entries from the logging bucket similar #1411, but couldn’t. 

According to [this document](https://cloud.google.com/logging/docs/reference/v2/rest/v2/entries/list), specify views in `resourceNames` for entries from logging buckets, as an alternative to parent resources.

---

- [x] Make sure to open an issue as a [bug/issue](https://github.com/googleapis/nodejs-logging/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests and linter pass

Fixes #1411 🦕

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -593,9 +593,13 @@ class Logging {
     reqOpts.resourceNames = arrify(reqOpts.resourceNames!);
     this.projectId = await this.auth.getProjectId();
     const resourceName = 'projects/' + this.projectId;
-    if (reqOpts.resourceNames.indexOf(resourceName) === -1) {
-      reqOpts.resourceNames.push(resourceName);
-    }
+    const isReadFromBucket = reqOpts.resourceNames.some(resourceName =>
+      resourceName.startsWith(resourceName)
+    );
+    if (
+      reqOpts.resourceNames.indexOf(resourceName) === -1 &&
+      !isReadFromBucket
+    ) {
     delete reqOpts.autoPaginate;
     delete reqOpts.gaxOptions;
     let gaxOptions = extend(
--- a/test/index.ts
+++ b/test/index.ts
@@ -616,6 +616,30 @@ describe('Logging', () => {
       await logging.getEntries(options);
     });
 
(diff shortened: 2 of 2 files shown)


Colt45s:
https://cla.developers.google.com/ not working 😢

Colt45s:
Could you please review my pull request, @cindy-peng ?

vjerci:
+1 on needing this.

